PR #1302, opened 2021-08-17
VRE-1493 - Fix permission check for collection unpublishing action.

ewelinajozwik:
Fixes [VRE-1493](https://thehyve.atlassian.net/browse/VRE-1493)


frankyhollywood (approved):
nice to have a dedicated role instead of implicit 'collection.accessMode === 'DataPublished' && isAdmin(currentUser)'

PR #1315, opened 2021-09-08
maximum pool size, so the client doesn't exceed server limits

frankyhollywood:


frankyhollywood:
wait until dev is stable

PR #1317, opened 2021-09-08
Feature/connectionpool size limit

frankyhollywood:


projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -51,6 +51,7 @@ public ViewStoreClientFactory(ViewsConfig viewsConfig, Config.ViewDatabase viewD
         databaseConfig.setPassword(viewDatabase.password);
         databaseConfig.setAutoCommit(false);
         databaseConfig.setConnectionTimeout(1000);
+        databaseConfig.setMaximumPoolSize(50);
```
ewelinajozwik, line 54: maybe later it's good to make this configurable (same way as the password), in case `50` is not optimal. For now should be fine!

ewelinajozwik (approved):

PR #1319, opened 2021-09-09
Add secure headers for all resources.

ewelinajozwik:
Apply relevant headers to all responses, following [OWASP Secure Headers Project guides](https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers).

Relates to [VRE-821](https://thehyve.atlassian.net/browse/VRE-821)

Remaining headers that were not added:

- `Strict-Transport-Security` - already added by default
- `Content-Security-Policy` - making strict policies here breaks the user interface, might be specific per installation

Headers that do not seem to bring any benefit (since we already have Pluto middle-layer with zuul taking care of filtering and routing rules):

- `Referrer-Policy`
- `Cross-Origin-Embedder-Policy`
- `Cross-Origin-Opener-Policy`
- `Cross-Origin-Resource-Policy`

### DIFF
--- a/projects/pluto/src/main/java/nl/fairspace/pluto/auth/filters/SecureResponseHeaderAppenderFilter.java
+++ b/projects/pluto/src/main/java/nl/fairspace/pluto/auth/filters/SecureResponseHeaderAppenderFilter.java
@@ -32,7 +32,7 @@ public class SecureResponseHeaderAppenderFilter implements Filter {
         httpServletResponse.setHeader("X-Permitted-Cross-Domain-Policies", "none");
         httpServletResponse.setHeader(
                 "Content-Security-Policy",
-                "default-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; font-src 'self'; style-src: 'self'"
+                "default-src 'self' 'unsafe-inline'; font-src 'self'; img-src 'self' 'unsafe-inline' https://*"
         );
         chain.doFilter(request, response);
     }
--- a/projects/pluto/src/main/java/nl/fairspace/pluto/auth/filters/SecureResponseHeaderAppenderFilter.java
+++ b/projects/pluto/src/main/java/nl/fairspace/pluto/auth/filters/SecureResponseHeaderAppenderFilter.java
@@ -1,0 +1,44 @@
+package nl.fairspace.pluto.auth.filters;
+
+import org.springframework.stereotype.Component;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * This filter adds secure headers to the response in order to increase the security of the application.
+ * Based on OWASP Secure Headers Project (https://owasp.org/www-project-secure-headers/).
+ *
+ * Relates to OWASP Top 10 2017, A6:2017-Security Misconfiguration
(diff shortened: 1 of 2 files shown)


gijskant:
For comparison, for the Palga portal we use:
- `Content-Security-Policy-Report-Only: default-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:`
- `X-XSS-Protection: 1; mode=block`

ewelinajozwik:
> `X-XSS-Protection: 1; mode=block`

The `X-XSS-Protection` is deprecated and the OWASP guide suggests to set it to 0: https://owasp.org/www-project-secure-headers/#x-xss-protection

The `Content-Security-Policy-Report-Only` was not on the list, only the `Content-Security-Policy`. I can check what is the difference. I tried a similar values for `Content-Security-Policy` and the UI was not showing up (at least locally), but I can still try to tune this a bit. I'm just afraid it depends on the installation and may break things.

ewelinajozwik:
I added `Content-Security-Policy`. From what I checked `Content-Security-Policy-Report-Only` only prints violation reports to the console and delivers them to a violation endpoint if the report-to and report-uri directives are used (we don't have such an endpoint), but it does not block violations.

I did not add a policy for image-src, because it was indeed blocking images in the markdown field.

gijskant:
> I did not add a policy for image-src, because it was indeed blocking images in the markdown field.

But then it uses the policy in `default-src` as fallback (which means rejecting in this case).

projects/pluto/src/main/java/nl/fairspace/pluto/auth/filters/SecureResponseHeaderAppenderFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    public void doFilter(ServletRequest request, ServletResponse response,
+                         FilterChain chain) throws IOException, ServletException {
+        HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+        httpServletResponse.setHeader("X-Frame-Options", "DENY");
+        httpServletResponse.setHeader("X-Content-Type-Options", "nosniff");
+        httpServletResponse.setHeader("X-XSS-Protection", "0");
+        httpServletResponse.setHeader("X-Permitted-Cross-Domain-Policies", "none");
+        httpServletResponse.setHeader(
+                "Content-Security-Policy",
+                "default-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; font-src 'self'; style-src: 'self'"
```
gijskant: - `script-src` and `style-src` can inherit from `default-src`
- Allow including images from all https addresses
```suggestion
                "default-src 'self' 'unsafe-inline'; font-src 'self'; img-src 'self' 'unsafe-inline' https://*"
```

ewelinajozwik: Thanks for the suggestion! I applied it.

gijskant (approved):
Advised headers applied. Application is still working as expected. External images, referenced in Markdown fields are allowed, if https is used, otherwise images are blocked (as they should).
(When we add a description of the Markdown format somewhere, this could be mentioned.)

PR #1328, opened 2021-10-27
Clarify the data recovery documentation.

ewelinajozwik:


gijskant (approved):
I did not test the procedure, but the explanation is clear.

PR #1336, opened 2021-12-08
Fix facet selection issue.

ewelinajozwik:
Fixes: [VRE-1541](https://thehyve.atlassian.net/browse/VRE-1541)

### DIFF
--- a/projects/mercury/src/metadata/views/MetadataView.js
+++ b/projects/mercury/src/metadata/views/MetadataView.js
@@ -91,8 +91,8 @@ export const MetadataView = (props: MetadataViewProperties) => {
         if (filterCandidates.find(f => f.field === facet.name)) {
             let updatedFilters;
             const existingFilter = filters.find(f => f.field === facet.name);
-            if (!newValues || (newValues.filter(v => v !== null).length === 0) || (newValues && existingFilter && existingFilter.value
-                && _.isEqual(existingFilter.values.sort(), newValues.sort()))) {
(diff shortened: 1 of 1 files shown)


projects/mercury/src/metadata/views/MetadataView.js
```diff
@@ -91,8 +91,8 @@ export const MetadataView = (props: MetadataViewProperties) => {
         if (filterCandidates.find(f => f.field === facet.name)) {
             let updatedFilters;
             const existingFilter = filters.find(f => f.field === facet.name);
-            if (!newValues || (newValues.filter(v => v !== null).length === 0) || (newValues && existingFilter && existingFilter.value
-                && _.isEqual(existingFilter.values.sort(), newValues.sort()))) {
+            if (!newValues || (_.isEqual(existingFilter.values.sort(), newValues.sort()) && (
+                (newValues.filter(v => v !== null).length === 0) || (newValues && existingFilter && existingFilter.value)))) {
```
frankyhollywood: newValues is always true, otherwise !newValues from line 94 would be true

same for existingFilter, otherwise existingFilter.values.sort() would cause a null ref exception
```suggestion
                (newValues.filter(v => v !== null).length === 0) || existingFilter.value))) {
```

ewelinajozwik: Right, thanks @FrankyHollywood! I updated the condition here

frankyhollywood (approved):

PR #1358, opened 2022-07-08
Select metadata rows and export to csv

frankyhollywood:


projects/mercury/src/metadata/views/MetadataViewTableContainer.js
```diff
@@ -210,6 +253,8 @@ export const MetadataViewTableContainer = (props: MetadataViewTableContainerProp
                 history={history}
                 textFiltersObject={textFiltersObject}
                 setTextFiltersObject={setTextFiltersObject}
+                checkboxes={checkboxes}
+                setCheckboxState={setCheckboxState}
```
ewelinajozwik, line 262: There are some issues with reseting the selection, for example when you select some rows, then apply a filter that gives you 0 results, after exporting you'll still get the previously selected rows. I would limit the export only to the currently displayed rows. 
Also there is no overview of what you have currently selected... There should be some count at least and some overall checkbox that allows you to select/deselect everything from the current page or clear selection...
And I think when you go to a new page, the selection should be also cleared.

projects/mercury/src/metadata/views/MetadataViewTableContainer.js
```diff
@@ -227,18 +272,28 @@ export const MetadataViewTableContainer = (props: MetadataViewTableContainerProp
...
-                    ActionsComponent={TablePaginationActions}
-                />
+                <div className={classes.footerButtonDiv}>
+                    <Button
+                        className={classes.exportButton}
+                        onClick={exportTable}
+                        variant="contained"
+                        endIcon={<GetAppIcon fontSize="small" />}
+                    >
+                        Export
```
ewelinajozwik: I would name the button "Download selection", possibly with a count of selected rows, e.g. "Download selection (5)". The other button could be called them "Export to analysis environment" or something like this (preferably shorter)

projects/mercury/src/metadata/views/MetadataViewTableContainer.js
frankyhollywood: I agree :)

PR #1362, opened 2022-09-20
select all checkbox, and reset when data is updated

frankyhollywood:
Allow multi select in table, and an export button for metadata table

### DIFF
--- a/projects/mercury/src/metadata/views/MetadataViewTable.js
+++ b/projects/mercury/src/metadata/views/MetadataViewTable.js
@@ -173,7 +173,7 @@ export const MetadataViewTable = (props: MetadataViewTableProperties) => {
             <TableHead>
                 <TableRow>
                     <TableCell style={{padding: 0}}>
-                        <Checkbox id="checkAll" key={Math.random()} onChange={handleCheckallChange} defaultChecked={rowCount > 0 && checkedCount === rowCount} />
+                        <Checkbox id="checkAll" key={'checkall-key-' + checkedCount} onChange={handleCheckallChange} defaultChecked={rowCount > 0 && checkedCount === rowCount} />
                     </TableCell>
                     {visibleColumns.map(column => (
                         <TableCell key={column.name} className={classes.headerCellContents}>
--- a/projects/mercury/src/metadata/views/MetadataViewTable.js
+++ b/projects/mercury/src/metadata/views/MetadataViewTable.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import {Link, Table, TableBody, TableCell, TableHead, TableRow} from '@material-ui/core';
+import {Checkbox, Link, Table, TableBody, TableCell, TableHead, TableRow} from '@material-ui/core';
 import {makeStyles} from '@material-ui/core/styles';
 import {Link as RouterLink} from "react-router-dom";
 import type {MetadataViewColumn, MetadataViewData} from "./MetadataViewAPI";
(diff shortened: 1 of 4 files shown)


projects/mercury/src/metadata/views/MetadataViewTable.js
```diff
@@ -131,16 +154,23 @@ export const MetadataViewTable = (props: MetadataViewTableProperties) => {
...
+    const rowCount = data.rows && data.rows.length;
+
     return (
         <Table data-testid="results-table" size="small" stickyHeader={!loading}>
             <TableHead>
                 <TableRow>
+                    <TableCell style={{padding: 0}}>
+                        <Checkbox id="checkAll" key={Math.random()} onChange={handleCheckallChange} defaultChecked={rowCount > 0 && checkedCount === rowCount} />
+                    </TableCell>
+                    <TableCell />
```
ewelinajozwik: Why is this empty cell added in the header?
There is some issue with the table now. The headers are shifted by one column (header does not match the cells in the column, e.g. below subject header is above gender column):
Screenshot from 2022-09-23 12-42-05

Also there s something wrong with rows styling - the gray line below every row is now only covering part of the row.

projects/mercury/src/metadata/views/MetadataViewTableContainer.js
```diff
@@ -227,18 +317,28 @@ export const MetadataViewTableContainer = (props: MetadataViewTableContainerProp
...
-                    rowsPerPage={rowsPerPage}
-                    page={page}
-                    onPageChange={handleChangePage}
-                    onRowsPerPageChange={handleChangeRowsPerPage}
-                    className={classes.tableFooter}
-                    labelDisplayedRows={(d) => labelDisplayedRows({...d, countIsLoading: loadingCount})}
-                    ActionsComponent={TablePaginationActions}
-                />
+                <div className={classes.footerButtonDiv}>
+                    <Button
```
ewelinajozwik, line 321: I think the button would look nicer with extended width:
Screenshot from 2022-09-23 12-49-44

ewelinajozwik: And maybe it could be disabled if selection=0?

projects/mercury/src/metadata/views/MetadataViewTable.js
```diff
@@ -53,6 +53,17 @@ export const MetadataViewTable = (props: MetadataViewTableProperties) => {
...
+
+    // initialize checkboxes
+    if (idColumn && data && data.rows) {
+        data.rows.forEach(row => {
+            const key = row[idColumn.name][0].value;
+            if (checkboxes[key] === undefined) {
+                setCheckboxState(key, false);
+            }
+        });
+    }
```
ewelinajozwik: it would be better to wrap this check into useEffect hook with parameters, otherwise it keeps running this check and there are some rendering conflicts with other components (see browser console errors)

projects/mercury/src/metadata/views/MetadataViewTable.js
```diff
@@ -156,14 +186,14 @@ export const MetadataViewTable = (props: MetadataViewTableProperties) => {
                         key={row[idColumn.name][0].value}
                         hover={isResourcesView}
                         selected={selected && selected.iri === row[idColumn.name][0].value}
-                        onClick={() => handleResultSingleClick(
+                        onDoubleClick={() => handleResultDoubleClick(row[idColumn.name][0].value, row)}
+                    >
+                        <Checkbox id={row[idColumn.name][0].value} key={Math.random()} defaultChecked={checkboxes[row[idColumn.name][0].value]} onChange={handleCheckboxChange} />
```
ewelinajozwik, line 195: According to [React documentation](https://reactjs.org/docs/reconciliation.html) using Math.random is not a good way to generate a key:

> Keys should be stable, predictable, and unique. Unstable keys (like those produced by Math.random()) will cause many component instances and DOM nodes to be unnecessarily recreated, which can cause performance degradation and lost state in child components.

Maybe better to use same key as for the row (line 186):
`key={row[idColumn.name][0].value}`

projects/mercury/src/metadata/views/MetadataViewTableContainer.js
frankyhollywood: yes, that's a good idea!

projects/mercury/src/metadata/views/MetadataViewTable.js
frankyhollywood: yes this is a workaround, but it is on purpose. 

When the key is not random the checkall box doesn't always update correctly. For example, with page size 10 all checkboxes are checked. Now uncheck 1 checkbox so only 9 checkboxes are checked. Without using a random key the checkall is still checked which is incorrect. 

With using a random key this wrong behavior is not present. You have a suggestion how to solve this properly?

frankyhollywood: I made #checkedboxes part of the key. Still a workaround to force a control refresh, but better than random key

ewelinajozwik (approved):
@FrankyHollywood Nice, now it works well!

PR #1374, opened 2023-01-02
React 18

frankyhollywood:
- react 16 -> 18
- material ui 4 -> mui 5


projects/mercury/package.json
```diff
@@ -19,18 +19,18 @@
...
-    "@material-ui/pickers": "^3.2.8",
-    "@mui/material": "^5.5.3",
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
+    "@mui/icons-material": "^5.11.0",
+    "@mui/lab": "^5.0.0-alpha.113",
+    "@mui/material": "^5.11.1",
+    "@mui/styles": "^5.11.1",
+    "@mui/x-date-pickers": "^5.0.11",
     "axios": "^0.26.1",
```
ewelinajozwik: we're also far behind with axios, could you upgrade to 1.2.2 if possible?

ewelinajozwik:
Maybe we could also upgrade node version to v16.14.1 (LTS)?

ewelinajozwik (changes requested):
Looks like the tests are failing on " Cannot find module 'enzyme-adapter-react-16'"

projects/mercury/src/collections/CollectionInformationDrawer.js
```diff
@@ -203,14 +203,15 @@ const MetadataCard = (props) => {
                                         </>
                                     )}
                                 >
-                                    <IconButton onClick={open}><CloudUpload /></IconButton>
+                                    <IconButton onClick={open} size="large"><CloudUpload /></IconButton>
```
ewelinajozwik: why `large` instead of the default `medium`?

ewelinajozwik:
I would expand the margins, now it seems that's really dense in some components, e.g. adding new file:
Screenshot from 2023-01-06 08-30-52

Also I would extend the menu item margins and collection details panel as it was before:

Screenshot from 2023-01-06 08-52-02

ewelinajozwik:
There is also a bug when adding new users to workpaces - list of users to be added is empty, not sure if it was there before (it works well on fns-demo). Could you check if it works for you?

ewelinajozwik:
Could you also squash all the "material-ui 4 -> mui 5" commits into one at the end?

ewelinajozwik (approved):
I just did one last test - works well and looks good!

PR #1382, opened 2023-01-25
Upgraded to latest Keycloak

FemkeK:
- Upgraded to latest Keycloak
- Added realm export file for import from UI
- Fixed Saturn wait URL
- Updated OIDC URLs in application.yml to look for correct Keycloak data
- Confirmed compatibility with Apple M1 processor


local-development/.env
```diff
@@ -3,3 +3,4 @@ KEYCLOAK_REALM=fairspace
 WORKSPACE_CLIENT_ID=workspace-client
 WORKSPACE_URL=http://localhost:8080
 WORKSPACE_DEBUG_URL=http://localhost:3000
+DOCKER_LOGGING_DRIVER=json-file
```
ewelinajozwik: this is needed for mac users, linux users should use the default journald, so this shouldn't be added in the default configuration (only your local .env)

local-development/docker-compose.yml
```diff
@@ -20,22 +20,19 @@ services:
         tag: fairspace-keycloak-postgres
 
   fairspace-keycloak:
-    image: jboss/keycloak:12.0.4
+    image: keycloak/keycloak:latest
```
ewelinajozwik: better to specify the version, otherwise if the next one has some breaking changes, the default setup will stop working
you were testing with 20.0.3, right?

projects/mercury/package.json
```diff
@@ -57,7 +57,8 @@
     "text-table": "^0.2.0",
     "typeface-roboto": "^1.1.13",
     "use-deep-compare-effect": "^1.3.0",
-    "webdav": "^4.0.0"
+    "webdav": "^4.0.0",
+    "webpack": "^5.75.0"
```
ewelinajozwik: webpack is a dependency of e.g. react-scripts, so it will be fetched anyway and there might be some version conflicts now. It shouldn't be necessary to specify it here, unless you had some issues with this? If so, do you remember what was the problem?

projects/pluto/src/main/resources/application.yml
```diff
@@ -45,10 +45,10 @@ logging:
...
 security:
   oidc:
-    auth-url: ${pluto.oauth2.base-url}/realms/${pluto.oauth2.realm}/protocol/openid-connect/auth
-    logout-url: ${pluto.oauth2.base-url}/realms/${pluto.oauth2.realm}/protocol/openid-connect/logout?redirect_uri=%s
-    token-url: ${pluto.oauth2.backend-url}/realms/${pluto.oauth2.realm}/protocol/openid-connect/token
-    jwk-key-set-url: ${pluto.oauth2.backend-url}/realms/${pluto.oauth2.realm}/protocol/openid-connect/certs
+    auth-url: ${pluto.oauth2.base-url}/auth/realms/${pluto.oauth2.realm}/protocol/openid-connect/auth
+    logout-url: ${pluto.oauth2.base-url}/auth/realms/${pluto.oauth2.realm}/protocol/openid-connect/logout?redirect_uri=%s
+    token-url: ${pluto.oauth2.backend-url}/auth/realms/${pluto.oauth2.realm}/protocol/openid-connect/token
+    jwk-key-set-url: ${pluto.oauth2.backend-url}/auth/realms/${pluto.oauth2.realm}/protocol/openid-connect/certs
```
ewelinajozwik: the "/auth" part was removed from the OIDC URL in Keycloak 18. For the setup we have for fns and KEycloak 20 removing `/auth` part was necessary... Why this change to add it back?

ewelinajozwik (commented):
I will test if this setup works for me locally

local-development/.env
FemkeK: Ah yes, I meant to leave it out of the commit. This should indeed not be added to the default configuration.

local-development/docker-compose.yml
FemkeK: Yes, indeed I tested with 20.0.3👍🏻

projects/mercury/package.json
FemkeK: This was automatically updated, I didn't do it manually.

projects/pluto/src/main/resources/application.yml
FemkeK: I added it back because otherwise I receive `Exception in thread "main" java.lang.reflect.InvocationTargetException` exceptions, causing the fairspace-keycloak container to repeatedly close and restart again.

local-development/docker-compose.yml
frankyhollywood: Well let's discuss during standup, I actually like it when it breaks so we can fix it. We should try to keep up to date with tooling, especially with the frontend security barrier. It will break locally first, not at a customer. If somehow we can't fix a breaking change we can always make it fixed.

projects/pluto/src/main/resources/application.yml
ewelinajozwik: you're right, also for me locally it still requires /auth... That is strange, because according to the documentation it shouldn't be the case. 
Anyway we need to enable setting these parameters without /auth, otherwise fns won't work... 
Let's discuss tomorrow

local-development/keycloak/realm-export.json
```diff
@@ -0,0 +1,1969 @@
...
+      "failureFactor": "30",
+      "actionTokenGeneratedByUserLifespan": "300",
+      "maxDeltaTimeSeconds": "43200",
+      "_browser_header.xContentTypeOptions": "nosniff",
+      "actionTokenGeneratedByAdminLifespan": "43200",
+      "bruteForceProtected": "false",
+      "_browser_header.contentSecurityPolicy": "frame-src 'self'; frame-ancestors 'self'; object-src 'none';",
+      "waitIncrementSeconds": "60"
+    },
+    "keycloakVersion": "3.4.3.Final"
```
ewelinajozwik: ?

projects/mercury/package.json
ewelinajozwik: Could you remove it? It must have been added by an install command that you ran locally.

projects/mercury/yarn.lock
ewelinajozwik, line 1: Please don't change this file, unless you're adding some new dependencies (not the case in this PR)

local-development/docker-compose.yml
```diff
@@ -44,14 +41,17 @@ services:
       - fairspace-keycloak-db-network
     volumes:
       - ./keycloak/setup-realm.sh:/opt/jboss/startup-scripts/setup-realm.sh
-      - ./keycloak/realm-template.json:/tmp/realm-template.json
-      - ./keycloak/keycloak-add-user.json:/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json
+      # - ./keycloak/realm-template.json:/tmp/realm-template.json
+      - ./keycloak/realm-export.json:/opt/keycloak/data/import/realm-export.json
```
ewelinajozwik: so the realm-export.json replaces realm-template.json? 
I see that now we have both, but only one is used

local-development/keycloak/realm-export.json
```diff
@@ -0,0 +1,1969 @@
...
+            }
+          }
+        ],
+        "useTemplateConfig": false,
+        "useTemplateScope": false,
+        "useTemplateMappers": false
+      },
+      {
+        "id": "88e1ea97-dbd0-41f7-801f-a99404f0c11b",
+        "clientId": "workspace-client",
```
ewelinajozwik: It looks like you hardcoded the client id and realm name (line 3), before (in realm-template.json) it was configurable based on env variables.

local-development/docker-compose.yml
```diff
@@ -20,22 +20,19 @@ services:
...
     environment:
-      DB_VENDOR: POSTGRES
-      DB_ADDR: fairspace-keycloak-postgres
-      DB_DATABASE: keycloak
-      DB_USER: keycloak
-      DB_PASSWORD: keycloak
-      KEYCLOAK_HOSTNAME: ${KEYCLOAK_HOSTNAME:?Please configure KEYCLOAK_HOSTNAME.}
-      KEYCLOAK_HTTP_PORT: ${KEYCLOAK_PORT:-5100}
-      KEYCLOAK_IMPORT: /tmp/realm-export.json
-      KEYCLOAK_REALM: ${KEYCLOAK_REALM:?Please configure KEYCLOAK_REALM}
```
ewelinajozwik, line 34: this one was used mainly in the realm template, currently it's not set at all...

ewelinajozwik (approved):

PR #1383, opened 2023-01-25
allow for default roles at initialisation of Fairspace

frankyhollywood:


projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
```diff
@@ -68,6 +68,7 @@ public static class Auth {
         public String clientId = "workspace-client";
         public boolean enableBasicAuth;
         public String superAdminUser = "organisation-admin";
+        public String[] defaultRoles = new String[0];
```
ewelinajozwik: wouldn't it be better to have it as set?
```
    @JsonSetter(nulls = Nulls.AS_EMPTY)
    public final Set<String> defaultRoles = new HashSet<>();
```

projects/saturn/application.yaml
```diff
@@ -42,6 +42,8 @@ auth:
   clientId: workspace-client
   enableBasicAuth: true
   superAdminUser: organisation-admin
+  defaultRoles:
+    - canViewPublicData
```
ewelinajozwik: I wouldn't make this one as default. For fns we need `canViewPublicMetadata` instead, but maybe anyway it's more safe to have it empty by default and have it configurable by helm in configmap-saturn.yaml to set it only for fns deployment

frankyhollywood: for now we can have it canViewPublicMetadata , maybe change with Synergy if not needed. I guess most customers want view public metadata on by default, it's public :)

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
frankyhollywood: I don't know, isn't a HashSet a bit of overkill for max 4 values? I would hash only bigger sets.

And also, will it serialize automatically? I don't like to write a custom serializer for this when this works OOTB.

projects/saturn/application.yaml
ewelinajozwik: but in the PR you made canView**PublicData** default, not canView**PublicMetadata**

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
ewelinajozwik: yes, it should work well, also it would made it consistent with other settings, e.g. features.
For `new String[0]` - have you checked what happens, when you define defaultRoles as null or empty? 
I'm not 100% sure, but I think we used to have some problems with properties defined this way...

frankyhollywood: ok I'll change it!

frankyhollywood: I will also change the name 'defaultRoles' to 'defaultUserRoles' because it is already used in the realm template, it might cause confusion

projects/saturn/application.yaml
frankyhollywood: yes that was a mistake, good you spot it :)

PR #1388, opened 2023-02-13
Validation in datetime facet

frankyhollywood:
- on keyboard entry, don't validate while typing
- on validation, don't clear the begin-date if the end-date is not correct
- do less validation, the MUI control validates for valid begin- and end-date ootb

### DIFF
--- a/projects/mercury/src/metadata/views/facets/DateSelectionFacet.js
+++ b/projects/mercury/src/metadata/views/facets/DateSelectionFacet.js
@@ -72,10 +72,10 @@ const DateSelectionFacet = (props: MetadataViewFacetProperties) => {
     return (
         <Grid container>
             <Grid item>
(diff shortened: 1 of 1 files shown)


projects/mercury/src/metadata/views/facets/DateSelectionFacet.js
```diff
@@ -33,33 +33,23 @@ const DateSelectionFacet = (props: MetadataViewFacetProperties) => {
         }
     }, [activeFilterValues]);
 
-    const isValid = (val: Date | null): boolean => {
```
ewelinajozwik, line 36: I'm not sure if fully removing this check is a good solution, now I can type anything, like 99/99/9999 and it doesn't show it's invalid date (not in red)...

frankyhollywood: ah that's not good indeed, that needs some rework

PR #1393, opened 2023-03-02
Fix issues with extra storage permissions.

ewelinajozwik:
Fixes [FNS-75](https://thehyve.atlassian.net/browse/FNS-75)

[FNS-75]: https://thehyve.atlassian.net/browse/FNS-75?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
@@ -60,6 +60,9 @@ public class DavFactory implements ResourceFactory {
     }
 
     public Resource getResource(org.apache.jena.rdf.model.Resource subject, Access access) {
+        if (isExtraStoreResource() && subject.hasProperty(RDF.type, FS.File) && !subject.hasProperty(FS.createdBy, currentUserResource())) {
+            return null;
+        }
         if (subject.hasProperty(FS.dateDeleted) && !showDeleted()) {
             return null;
         }
@@ -80,14 +83,18 @@ public class DavFactory implements ResourceFactory {
             return new CollectionResource(this, subject, access);
         }
         if (subject.hasProperty(RDF.type, FS.ExtraStorageDirectory)) {
-            return new DirectoryResource(this, subject, Access.Manage);
+            return new DirectoryResource(this, subject, access);
         }
 
         return null;
     }
 
+    public boolean isExtraStoreResource() {
+        return root instanceof ExtraStorageRootResource;
+    }
+
     public Access getAccess(org.apache.jena.rdf.model.Resource subject) {
-        if(root instanceof ExtraStorageRootResource) {
+        if (isExtraStoreResource()) {
             return getExtraStorageAccess(subject);
         }
 
@@ -104,7 +111,7 @@ public class DavFactory implements ResourceFactory {
 
         var access = getGrantedPermission(coll, user);
 
-        if(user.hasProperty(FS.isManagerOf, ownerWs)) {
+        if (user.hasProperty(FS.isManagerOf, ownerWs)) {
             access = Access.Manage;
         }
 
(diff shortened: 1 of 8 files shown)


projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
```diff
@@ -60,6 +60,9 @@ public Resource getResource(org.apache.jena.rdf.model.Resource subject) {
     }
 
     public Resource getResource(org.apache.jena.rdf.model.Resource subject, Access access) {
+        if (isExtraStoreResource() && subject.hasProperty(RDF.type, FS.File) && !subject.hasProperty(FS.createdBy, currentUserResource())) {
```
frankyhollywood, line 63: Having 'subject.hasProperty(RDF.type, FS.File)' means that a non-file resource (directory, collection) can be accessed by everyone.

projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
```diff
@@ -136,18 +143,19 @@ public Access getAccess(org.apache.jena.rdf.model.Resource subject) {
...
             return Access.List;
         }
 
         return access;
     }
 
     private Access getExtraStorageAccess(org.apache.jena.rdf.model.Resource subject) {
-        if (subject.hasProperty(FS.createdBy, currentUserResource()) && subject.hasProperty(RDF.type, FS.File)) {
-            return Access.Manage;
+        if (subject.equals(this.rootSubject)) {
```
frankyhollywood, line 154: could the order of if statements have a result that you get read access, while you should have write access?

projects/saturn/src/main/java/io/fairspace/saturn/webdav/resources/BaseResource.java
```diff
@@ -94,8 +94,19 @@ public Date getCreateDate() {
...
+        boolean purge;
+        if (factory.isExtraStoreResource()) {
+            if (subject.hasProperty(RDF.type, FS.ExtraStorageDirectory)) {
+                throw new NotAuthorizedException("Not authorized to purge the extra store root directory.", this, SC_FORBIDDEN);
+            }
+            purge = true;
+        } else {
+            purge = subject.hasProperty(FS.dateDeleted);
+            if (purge && !factory.userService.currentUser().isAdmin()) {
+                throw new NotAuthorizedException("Not authorized to purge the resource.", this, SC_FORBIDDEN);
```
frankyhollywood, line 107: why does the non-extraStorageResource path an extra validation + exception, was the old good not suffictient?

frankyhollywood (approved):

projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
ewelinajozwik: yes, I though for now there will be simple file-based permissions , as this is what we need for jupyter. In the future we can make extra storage permissions more complex (e.g. when we find another usecase for this storage).

projects/saturn/src/main/java/io/fairspace/saturn/webdav/DavFactory.java
ewelinajozwik: yes, the order matters, this is to protect the root folder being deleted by a user (it is created on application startup and should not be deleted), so it has to be read-only.

projects/saturn/src/main/java/io/fairspace/saturn/webdav/resources/BaseResource.java
ewelinajozwik: the validation was there, but I just moved it from `protected void delete(boolean purge)` to `public final void delete()` function, so it is done earlier on and so I can overwrite the protected one in other classes without repeating the validation in all the places.

PR #1395, opened 2023-03-08
Metadataview performance

frankyhollywood:
* Improved performance by windowed rendering, and splitting MetadataView in smaller components which prevents unnecessary rendering.

* Some small GUI changes


projects/mercury/src/metadata/views/ViewFacets.js
```diff
@@ -0,0 +1,130 @@
...
+    facetHeaders: {
+        textAlign: 'left',
+        marginTop: 6,
+        fontSize: 13,
+        color: theme.palette.primary.light,
+        marginLeft: 4
+    }
+});
+
+export const ViewFacets = (props) => {
```
ewelinajozwik: Could you change it to "MetadataViewFacets" to keep the naming consistent?

projects/mercury/src/metadata/views/ViewTabs.js
```diff
@@ -0,0 +1,80 @@
...
+        paddingRight: 70
+    },
+    tab: {
+        '& .MuiBox-root': {
+            padding: 0,
+        },
+    }
+});
+
+export const ViewTabs = (props) => {
```
ewelinajozwik: Same here, could you rename to "MetadataViewTabs"?

ewelinajozwik: Could you also define the type for props, like it's done for existing metadata view classes?

projects/mercury/src/metadata/views/ViewFacets.js
ewelinajozwik: Could you also define the type for props, like it's done for existing metadata view classes?

projects/mercury/src/metadata/views/facets/TextSelectionFacet.js
```diff
@@ -99,7 +106,17 @@ const SelectMultiple = (props: SelectProperties) => {
...
+            <List
+                height={150}
+                itemCount={filteredOptions.length}
+                itemSize={35}
+                width={300}
+            >
+                {renderCheckboxListElementWindowed}
+            </List>
+        );
+        // filteredOptions.map(option => renderCheckboxListElement(option));
```
ewelinajozwik: It looks like this line can be removed now.

projects/mercury/src/metadata/views/MetadataView.js
```diff
@@ -311,13 +178,31 @@ export const MetadataView = (props: MetadataViewProperties) => {
...
+                                updateFilterCandidates={updateFilterCandidates}
+                                handleClearFilter={handleClearFilter}
+                                applyFilters={applyFilters}
+                            />
                         </Grid>
                         <Grid item className={classes.metadataViewTabs}>
-                            {renderViewTabs()}
+                            <ViewTabs
+                                currentViewIndex={currentViewIndex}
+                                changeTab={changeTab}
```
ewelinajozwik, line 195: you are passing a lot of properties and functions to the new sub-components, while the functions can be defined in those subcomponents instead. 
I think that MetadataView.js should now only contain the shared functions/properties between both views and filters, e.g. changeTab is only relevant in ViewTabs, so why define it here and pass, instead of defining in ViewTabs?

ewelinajozwik (changes requested):
I cannot run this branch, getting 

> ERROR in ./src/metadata/views/facets/TextSelectionFacet.js 15:0-53
Module not found: Error: Can't resolve 'react-window' in .../projects/mercury/src/metadata/views/facets'

ewelinajozwik:
Also, the tests are failing, probably same "Can't resolve 'react-window'" error

ewelinajozwik:
Also the styling of facets is now a bit weird - there is a lot of empty space and always a horizontal scroll included:
Screenshot from 2023-03-09 15-42-53

projects/mercury/src/metadata/views/MetadataViewTableContainer.js
```diff
@@ -353,28 +359,35 @@ export const MetadataViewTableContainer = (props: MetadataViewTableContainerProp
...
-                        endIcon={<GetAppIcon fontSize="small" />}
-                        disabled={checkedCount === 0}
-                    >
-                        Download selection ({checkedCount})
-                    </Button>
-                    {exportToAnalysisEnabled && (
-                        <ProgressButton active={exportToAnalysisLoading}>
-                            <Button
+                    <div className={classes.footerCountDiv}>
+                        export items: {checkedCount}
```
ewelinajozwik: Or "Selected rows:"?
Could you use Typography component here? I thing it would look better if this text and "Rows per page:" text that is next to it have the same styling (color, size, font):
Screenshot from 2023-03-09 15-43-43

ewelinajozwik: and when the left margin is the same as for the checkboxes above

projects/mercury/src/metadata/views/ViewFacets.js
frankyhollywood: > to keep the naming consistent
I did it on purpose, because these two are sub-components which are now nicely grouped, but I agree with the consistency argument, will change it

projects/mercury/src/metadata/views/facets/TextSelectionFacet.js
frankyhollywood: looks like you're right :D

projects/mercury/src/metadata/views/MetadataView.js
frankyhollywood: Within changeTab other functions are called which are relevant for MetadataView

ewelinajozwik (approved):

PR #1402, opened 2023-03-21
prevent double scroll when table exeeds max height

frankyhollywood:
Table paging controls were not visible for large tables, especially when filters are active. Fixed by styling finetune

### DIFF
--- a/projects/mercury/src/metadata/views/MetadataViewTable.js
+++ b/projects/mercury/src/metadata/views/MetadataViewTable.js
@@ -15,6 +15,7 @@ import ColumnFilterInput from "../../common/components/ColumnFilterInput";
 
 type MetadataViewTableProperties = {
     data: MetadataViewData;
+    loading: boolean;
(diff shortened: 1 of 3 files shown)


projects/mercury/src/metadata/views/MetadataViewTable.js
```diff
@@ -48,7 +47,7 @@ const CUSTOM_RESOURCE_COLUMNS = ['access', 'path'];
 const RESOURCE_TYPE_COLUMN = `${RESOURCES_VIEW}_type`;
 
 export const MetadataViewTable = (props: MetadataViewTableProperties) => {
-    const {columns, visibleColumnNames, loading, data, toggleRow, selected, view, idColumn, history, collections} = props;
```
ewelinajozwik: the `loading` flag is still passed as a property to this component from `MetadataViewTableContainer` and in tests, could you remove it there?

ewelinajozwik (approved):

PR #1409, opened 2023-04-19
handle large lists in metadata panel

frankyhollywood:
- windowed rendering,
- prevent slow item lookup
- show number of items in list

### DIFF
--- a/projects/mercury/src/metadata/common/LinkedDataProperty.js
+++ b/projects/mercury/src/metadata/common/LinkedDataProperty.js
@@ -56,7 +56,7 @@ const LinkedDataProperty = (
                     ? (
                         <GenericTooltip interactive leaveDelay={100} title={labelTooltip} enterDelay={TOOLTIP_ENTER_DELAY}>
                             <FormLabel component="legend">
-                                {label}
+                                {label} {values.length > 1 && "(" + values.length + ")"}
                             </FormLabel>
                         </GenericTooltip>
                     ) : null
--- a/projects/mercury/src/metadata/common/jsonLdConverter.js
+++ b/projects/mercury/src/metadata/common/jsonLdConverter.js
@@ -7,13 +7,12 @@ import {compareBy, comparing, flattenShallow, isNonEmptyValue} from "../../commo
 /**
  * Generates an entry to describe a single value for a property
  * @param entry
- * @param allMetadata
  * @returns {{id: *, label, value: *}}
  */
-const generateValueEntry = (entry, allMetadata) => ({
+const generateValueEntry = (entry) => ({
     id: entry['@id'],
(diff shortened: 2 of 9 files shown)


projects/mercury/src/metadata/common/jsonLdConverter.js
frankyhollywood, line 1: Main change is some restructuring, made it easier to find the performance bottleneck. Turned out to be the 'generateValueEntry' method. 

I do like to keep it like this, because the conditional is now more clear:

if (predicateUri.startsWith('_')) {
            processFirstPredicateId(predicateUri);
        } else {
            processValues(predicateUri);

ewelinajozwik: It looks like something doesn't work correctly now, because it displays id instead of label for some properties, see below:
Screenshot from 2023-04-25 15-09-28

projects/mercury/src/metadata/common/LinkedDataProperty.js
```diff
@@ -56,7 +56,7 @@ const LinkedDataProperty = (
                     ? (
                         <GenericTooltip interactive leaveDelay={100} title={labelTooltip} enterDelay={TOOLTIP_ENTER_DELAY}>
                             <FormLabel component="legend">
-                                {label}
+                                {label} {values.length > 1 && "(" + values.length + ")"}
```
ewelinajozwik, line 59: nice idea and looks good!

projects/mercury/src/metadata/common/jsonLdConverter.js
```diff
@@ -7,13 +7,12 @@ import {compareBy, comparing, flattenShallow, isNonEmptyValue} from "../../commo
...
  * @param entry
- * @param allMetadata
  * @returns {{id: *, label, value: *}}
  */
-const generateValueEntry = (entry, allMetadata) => ({
+const generateValueEntry = (entry) => ({
     id: entry['@id'],
     value: entry['@value'],
-    label: entry['@id'] && getLabel(allMetadata.find(element => element['@id'] === entry['@id']))
+    label: entry['@id'] && getLabel(entry)
```
ewelinajozwik: it looks like this is the part where it doesn't work now

ewelinajozwik (changes requested):
The list looks really good, but there is an issue with the labels and there are some failing tests.

projects/mercury/src/metadata/common/LinkedDataProperty.js
frankyhollywood: thx :)

ewelinajozwik (approved):
I tested and it works fine now!

PR #1413, opened 2023-06-06
Npm update

frankyhollywood:
https://thehyve.atlassian.net/browse/FNS-121
https://thehyve.atlassian.net/browse/FNS-120

- Postgres 13 -> 15
- Node 16 -> 18
- Alpine 3.9 -> 3.18

### DIFF
--- a/local-development/docker-compose.yml
+++ b/local-development/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   fairspace-keycloak-postgres:
-    image: postgres:13-alpine
+    image: postgres:15-alpine
     container_name: fairspace-keycloak-database
     environment:
       POSTGRES_DB: keycloak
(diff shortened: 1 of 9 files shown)


projects/mercury/jest-setup.js
ewelinajozwik, line 1: Isn't this file needed only for your local setup?

local-development/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   fairspace-keycloak-postgres:
-    image: postgres:13-alpine
```
ewelinagr, line 5: Could you also change it in Keycloak repo for consistency?

ewelinajozwik (commented):
I tested with minikube, all works well, although we'll need to drop the database for deployment. 
Only one comment above.

projects/mercury/jest-setup.js
frankyhollywood: No it isn't, polyfills were removed from Node 18 so you need to provide global TextEncodere/decoder to make tests run.

it's in the comment

// TextEncoder and TextDecoder are not available anymore in the jest jsdom environment. They
// are available in all common browsers, therefor we can use them in tests without risk.
// https://github.com/inrupt/solid-client-authn-js/issues/1676

ewelinajozwik (approved):

PR #1415, opened 2023-06-15
Migration to typescript

frankyhollywood:


ewelinajozwik:
Build is failing due to linting errors

projects/mercury/.eslintrc.json
ewelinajozwik, line 1: why is this one removed?

projects/mercury/README.md
```diff
@@ -1,87 +1,46 @@
-# Mercury
+# Getting Started with Create React App
```
ewelinajozwik, line 1: You overwrited the original documentation, could you revert this change?

ewelinajozwik (changes requested):
Why did you changed some .js files to .tsx?

Also, you applied some weird formatting to almost all the changed files. Because of that most of them are not recognized here as renaming change, but as deletion of one file and adding a new one with .ts/.tsx extension.
Look at the difference in length between old and new files.

projects/mercury/.eslintrc.json
frankyhollywood: I have added it again

projects/mercury/README.md
frankyhollywood: yes :)

PR #1426, opened 2023-08-08
Fix folder upload issue

frankyhollywood:
We use the react-dropzone component. 

It doesn't support a folder dialog out of the box, only a workaround with input element see https://github.com/react-dropzone/react-dropzone/discussions/1157

In the existing solution a variable 'isFolderUpload' was used. The problem with this was the input element was already rendered, this caused a delay of 1 click. So if you click 'upload folder' you see a file dialog, click 'upload folder' again, now it is a folder dialog, because the variable was set by the previous click, and the second rendering uses this value. 

Previous solution tried to overcome this with this:
```

    // A hook to make sure that isFolderUpload state is changed before opening the upload dialog
    useEffect(() => {
        if (isFolderUpload !== undefined) {
            open();
        }
    }, [isFolderUpload, open]);
```
In Chrome and Edge this causes infinite re-rendering. Firefox blocks it. 

Solution for now is to remove the useEffect, en render two input elements, one for folders one for files.

### DIFF
--- a/projects/mercury/src/file/FileBrowser.js
+++ b/projects/mercury/src/file/FileBrowser.js
@@ -108,7 +108,7 @@ export const FileBrowser = (props: FileBrowserProperties) => {
     const [overwriteFolderCandidateNames, setOverwriteFolderCandidateNames] = useState([]);
     const [currentUpload, setCurrentUpload] = useState({});
 
-    const InitializeDropzone = () => useDropzone({
+    const useFairspaceDropzone = () => useDropzone({
         noClick: true,
         noKeyboard: true,
         multiple: true,
@@ -137,8 +137,8 @@ export const FileBrowser = (props: FileBrowserProperties) => {
         }
     });
 
-    const fileDropzoneProperties = InitializeDropzone();
-    const folderDropzoneProperties = InitializeDropzone();
+    const fileDropzoneProperties = useFairspaceDropzone();
+    const folderDropzoneProperties = useFairspaceDropzone();
 
     // Deselect all files on history changes
     useEffect(() => {
--- a/projects/mercury/src/file/FileBrowser.js
+++ b/projects/mercury/src/file/FileBrowser.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import React, {useContext, useEffect, useRef, useState} from 'react';
 import {withRouter} from "react-router-dom";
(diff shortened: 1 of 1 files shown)


projects/mercury/src/file/FileBrowser.js
```diff
@@ -107,42 +107,40 @@ export const FileBrowser = (props: FileBrowserProperties) => {
...
-        multiple: true,
-        onDropAccepted: (droppedFiles) => {
-            const newUpload = {
-                id: generateUuid(),
-                files: droppedFiles,
-                destinationPath: openedPath,
-            };
-            const newOverwriteFolderCandidates = getConflictingFolders(droppedFiles, existingFolderNames);
-            const newOverwriteFileCandidates = getConflictingFiles(droppedFiles, existingFileNames);
+    function InitializeDropzone() {
```
ewelinajozwik:
```suggestion
    const initializeDropzone = () => {
```
Just for consistency with the way we define functions in other places

projects/mercury/src/file/FileBrowser.js
```diff
@@ -107,42 +107,40 @@ export const FileBrowser = (props: FileBrowserProperties) => {
...
-            } else {
-                startUpload(newUpload).then(refreshFiles);
             }
-        }
-    });
+        });
+    }
+
+    const fileDropzoneProperties = InitializeDropzone();
+    const folderDropzoneProperties = InitializeDropzone();
```
ewelinajozwik: with single `dropZone = InitializeDropzone()` instead of separate fileDropZone and folderDropZone it also seems to work correctly

frankyhollywood: I tried that, but in the end didn't succeed because how would you handle the 2 click events? 

```
    const uploadFolder = () => {
        folderDropzoneProperties.open();
    };

    const uploadFile = () => {
        fileDropzoneProperties.open();
    };
```

ewelinajozwik: ok, you're right, I didn't notice that now there are always 2 input fields being rendered, no longer depending on the flag `isFolderUpload`

ewelinajozwik (commented):
Looks good now! Only one small comment about lowercasing `InitializeDropzone` function

ewelinajozwik (approved):

PR #1429, opened 2023-08-09
Increase checkout time to prevent pipeline run errors

frankyhollywood:
https://postgresqlco.nf/doc/en/param/checkpoint_timeout/

### DIFF
--- a/charts/fairspace/templates/project/stateful-set.yaml
+++ b/charts/fairspace/templates/project/stateful-set.yaml
@@ -25,7 +25,7 @@ spec:
       containers:
         - name: {{ .Release.Name }}-postgres-database
           image: postgres:15-alpine
-          args: ["postgres", "-c", "checkpoint_timeout=43200"]
+          args: ["-c", "checkpoint_timeout=43200"]
           volumeMounts:
             - name: postgres
               mountPath: /var/lib/postgresql/data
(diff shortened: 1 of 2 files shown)


local-development/docker-compose.yml
```diff
@@ -64,6 +64,7 @@ services:
     volumes:
       - fairspace-postgres-data:/var/lib/postgresql/data
     restart: unless-stopped
+    command: ["postgres", "-c", "checkpoint_timeout=43200"]
```
ewelinajozwik, line 67: this will work only for the local development, could you apply it also for Kubernetes deployment?

charts/fairspace/templates/project/stateful-set.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       containers:
         - name: {{ .Release.Name }}-postgres-database
           image: postgres:15-alpine
+          args: ["postgres", "-c", "checkpoint_timeout=43200"]
```
ewelinajozwik: Did you test it? I am not sure if this would work... 
In k8s configuration file you have a `command` field to define a command and `args` to define arguments for the command and here you use args to define both...  
So I think it should be `args: ["-c", "checkpoint_timeout=43200"]` or something like this

frankyhollywood: Yes I tested :)

Actually the 'postgres' command is not needed, because it is the default value see https://github.com/docker-library/postgres/blob/1a73ab671b5f0f18313726e734c76bf171385c32/15/bookworm/Dockerfile#L219

I will remove it

ewelinajozwik (approved):

PR #1433, opened 2023-08-28
Apply Pluto config change in Helm setup.

ewelinajozwik:


charts/fairspace/templates/project/configmap-pluto.yaml
```diff
@@ -68,6 +68,13 @@ data:
           default-filters:
             - RemoveRequestHeader=Pragma X-Frame-Options X-Content-Type-Options X-XSS-Protection X-Permitted-Cross-Domain-Policies Origin
             # Including the ORIGIN header would trigger CORS filtering downstream, but Pluto is already doing the filtering.
+            - name: Retry
```
frankyhollywood, line 71: Why do we need a Retry filter? I would expect our server to be responsive, and if that is not the case in general we should scale our backend, not have a proxy which sends retries.

ewelinajozwik: This change was already added in https://github.com/thehyve/fairspace/pull/1427 that you approved ;) 
It was a fix for premature connection closing (https://github.com/spring-cloud/spring-cloud-gateway/issues/2046#issuecomment-1561019873). 
I just forgot to apply it in the Helm config.

frankyhollywood: True but it still feels like a workaround, this need for retry. Is there a good reason for it, or it just works?

https://maxime-guilbert.medium.com/spring-gateway-connection-prematurely-closed-9ff10b7152f4

ewelinajozwik: yes, it is basically a workaround for a spring cloud gateway issue that many people struggle with (we did not have this issue with Zuul) and the cause is hard to identify. The configured retry is done only on this one single premature exception, not for all request related errors.

frankyhollywood (approved):

PR #1437, opened 2023-11-27
Limit number of items frontend receives

frankyhollywood:


projects/saturn/src/main/java/io/fairspace/saturn/services/users/UserService.java
```diff
@@ -81,7 +81,7 @@ public Map<Node, User> getUsersMap() {
     }
 
     private Map<Node, User> fetchUsers() {
-        var userCount = usersResource.count(); // TODO [FNS-126] implement proper pagination based on client request parameters
```
ewelinajozwik, line 84: this is still a TODO though (https://thehyve.atlassian.net/browse/FAIRSPC-7)

projects/saturn/src/main/java/io/fairspace/saturn/services/metadata/MetadataService.java
```diff
@@ -72,7 +72,9 @@ public Model get(String subject, boolean withValueProperties) {
                     .forEach(property -> m.listStatements(null, property, resource)
                             .filterKeep(stmt -> permissions.canReadMetadata(stmt.getSubject()))
                             .filterDrop(stmt -> stmt.getSubject().hasProperty(FS.dateDeleted))
-                            .forEachRemaining(stmt -> {
+                            .toList()
+                            .subList(0, 21)
```
ewelinajozwik: this doesn't seem like a performance fix - it will still need to fetch them all, transform to list and only then take a sublist...
Isn't there some Jena way to limit number of results on iterator level? I see there is a `limit()` method on stream\<Resource>, not sure how to do for stream\<Property> when filtering

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -388,13 +388,20 @@ List<Map<String, Set<ValueDTO>>> retrieveJoinTableRows(
...
+        // column filters.
         var query = connection.prepareStatement(
                 "select " + projectionColumns.stream().map(String::toLowerCase).collect(Collectors.joining(", ")) +
                         " from " + joinedTable.name + " j " +
                         " where exists (" +
                         "   select * from " + joinTable.name + " jt " +
                         "   where jt." + idColumn(joinView.view).name + " = j.id" +
-                        "   and jt." + idColumn(view).name + " = ? )"
+                        "   and jt." + idColumn(view).name + " = ? )" +
+                        "limit 50"
```
ewelinajozwik: Could we make it a variable to make it easier to spot?

projects/saturn/src/main/java/io/fairspace/saturn/services/users/UserService.java
frankyhollywood: yes but I don't prefer to have a backlog in code, rather have this in Atlassian... Than we have overview, and the text will not get out-dated (like 'FNS-126 is now 'FAIRSP-7')

projects/saturn/src/main/java/io/fairspace/saturn/services/metadata/MetadataService.java
frankyhollywood: Well it is because the data is not send to the client, which takes most time. I can make it even faster indeed by not doing the 'toList'.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
frankyhollywood: you mean replace '50' with a more meaningful name?

ewelinajozwik: yeah, just like a constant so it is clear that this is the current setup

projects/saturn/src/main/java/io/fairspace/saturn/services/metadata/MetadataService.java
ewelinajozwik: would be nice if possible. But is there some Jena way that would allow replacing toList? I wanted to suggest something, but couldn't find it easily

ewelinajozwik: Also it looks like this change caused tests to fail...

PR #1438, opened 2023-11-29
Limit items rightpanel

frankyhollywood:


projects/mercury/src/metadata/common/LinkedDataValuesList.js
```diff
@@ -114,6 +114,29 @@ export const LinkedDataValuesList = (props: LinkedDataValuesListProps) => {
...
+                }
+            </div>
+            {showRowDividers && <Divider />}
+        </Grid>
+    ));
+
+    const renderListLimitMessage = (entry) => rowDecorator(entry, (
+        <Grid item className={classes.values}>
+            <div>
+                <Tooltip title={"If a view tab is defined for '" + property.label + "' you can retrieve all available values there."}>
```
ewelinajozwik: I think there should be a comma before "you can ...": `If..., you can ...`

And just a suggestion, feel free to skip it: maybe for non-technical users it is not clear how to "retrieve" values... (sounds like an action to do). Maybe better to change "retrieve" to "find all available values" or "see available values""?

ewelinajozwik (commented):
I haven't check how it looks with 100 values and scroll bar (don't have that much test data), but when limit is 2, it looks really nice!

projects/mercury/src/metadata/common/LinkedDataValuesList.js
frankyhollywood: ah good suggestion, write it like my grandmother would understand :D

PR #1446, opened 2023-12-13
Set column order in views.yaml

frankyhollywood:


projects/mercury/src/metadata/views/MetadataViewTabs.js
```diff
@@ -53,7 +53,15 @@ export const MetadataViewTabs = (props: MetadataViewTabsProperties) => {
                 accessColumn
             ];
         }
-        return view.columns;
+        return view.columns.sort((a, b) => {
```
tgreenwood: Could be also written in a slightly more laconic way (but I do not insist):
`return view.columns.sort((a, b) => a.displayIndex - b.displayIndex);`

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
```diff
@@ -97,6 +99,8 @@ public static class JoinView {
             public boolean reverse = false;
             @JsonSetter(nulls = Nulls.AS_EMPTY)
             public List<String> include;
+            // displayIndex determines the order of columns on the view page, for joinView it is the column displaying the related entity
+            public Integer displayIndex = 999;
```
tgreenwood: If you needed to initiate with just a big integer, I would suggest to use Integer.MAX_VALUE, it's kind of an unspoken rule. Otherwise, with another integer value it is more like a "magic number"

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -175,23 +175,23 @@ public List<ViewDTO> getViews() {
         return viewsConfig.views.stream()
                 .map(v -> {
                     var columns = new ArrayList<ColumnDTO>();
-                    columns.add(new ColumnDTO(v.name, v.itemName == null ? v.name : v.itemName, ColumnType.Identifier));
+                    columns.add(new ColumnDTO(v.name, v.itemName == null ? v.name : v.itemName, ColumnType.Identifier, 0));
```
tgreenwood: could you please make the 0 as a const with proper naming for better readability?

tgreenwood (commented):
added a few simple comments, could you please take a look?

projects/mercury/src/metadata/views/MetadataViewTabs.js
frankyhollywood: yes, good suggestion

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
frankyhollywood: yes I will change it

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
FrankyHollywood: I think this is a bit to much, since the variable isn't used elsewhere. Also there is the IDE assistance:

image

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -175,23 +175,27 @@ public List<ViewDTO> getViews() {
         return viewsConfig.views.stream()
                 .map(v -> {
                     var columns = new ArrayList<ColumnDTO>();
-                    columns.add(new ColumnDTO(v.name, v.itemName == null ? v.name : v.itemName, ColumnType.Identifier));
+
+                    // The entity labal is the first column displayed, if you want a column before this label, assign a negative displayIndex value in views.yaml
+                    int entityLabelIndex = 0;
```
tgreenwood, line 180: Sorry for confusion, I did not mean that, was talking not about method param name, but the value we pass there. What I mean is that when I was reading the code I did not get why we put 0 there. But if we put a const with proper naming, it would be clear. Constant on class level like this:

`private final static int ENTITY_LABEL_INDEX = 0` (or public, so could be used in tests for asserts if we do the check there.

It also can stay be a const on method level, but in this case I would suggest to add 'final'.

And here is the [link](https://rules.sonarsource.com/java/RSPEC-109/) describing the magic numbers, Sonar will fail once we switch it on

tgreenwood: if we had a unit test on the getViews method, it also could be explain a reader what is expected and why, but there is no one :( Let's skip it for now, I think we will get back to it once code coverage is switched on

PR #1456, opened 2023-12-20
Dialog for metadata details

frankyhollywood:
ROAR-254

### DIFF
--- a/projects/mercury/src/metadata/common/LinkedDataLink.js
+++ b/projects/mercury/src/metadata/common/LinkedDataLink.js
@@ -1,48 +1,32 @@
 import React, {useContext} from 'react';
 import * as PropTypes from "prop-types";
 import {Box, Modal, Tooltip} from '@mui/material';
+import withStyles from '@mui/styles/withStyles';
 import CloseIcon from '@mui/icons-material/Close';
 import LinkedDataEntityPage from "./LinkedDataEntityPage";
 import UserContext from '../../users/UserContext';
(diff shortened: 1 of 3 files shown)


projects/mercury/src/metadata/common/LinkedDataLink.js
```diff
@@ -1,28 +1,61 @@
...
-
 /**
  * Renders a link to the metadata editor.
  *
  * @param props
  * @constructor
  */
+
+const styleModalDialog = {
+    position: 'absolute',
```
tgreenwood: I see that we use '.style.js' files to define styles. What do you think if we move it to a dedicated one?

tgreenwood (approved):

projects/mercury/src/metadata/common/LinkedDataLink.js
```diff
@@ -1,28 +1,45 @@
...
- * Renders a link to the metadata editor.
- *
- * @param props
- * @constructor
+ * Renders a link to the metadata editor in a modal dialog when clicked.
  */
-const LinkedDataLink = ({uri, children}) => {
+
+const renderModal = (classes, open, handleClose, uri) => (
+    <Modal
```
ewelinajozwik, line 14: It is suggested to use Dialog rather than lower-level Modal construct (see [https://mui.com/material-ui/react-modal/#modal](https://mui.com/material-ui/react-modal/#modal)). This would also make the dialog more consistent with the existing code and you wouldn't need custom styling class.

PR #1465, opened 2024-01-16
reduced number of db calls for viewstorereader

frankyhollywood:


projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewRow.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class ViewRow {
+    private Map<String, Set<ValueDTO>> data;
+
+    public ViewRow() {
+        data = new HashMap<>();
+    }
+
+    // TODO, make obsolete by ViewStoreReader refactor
+    public Map<String, Set<ValueDTO>> getRawData() {
```
tgreenwood, line 16: I would suggest to return an unmodifiable map, otherwise I am not sure if we need this wrapper

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewRowCollection.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class ViewRowCollection {
+
+    private Hashtable<String, List<ViewRow>> data = new Hashtable<String, List<ViewRow>>();
+
+    public ViewRowCollection() {
+    }
+
+    public void add(String id, ViewRow row) {
+        if (!data.containsKey(id)) {
```
tgreenwood: `data.putIfAbsent(id, new ArrayList<>());` could be used instaed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewRowCollection.java
```diff
@@ -0,0 +1,27 @@
+package io.fairspace.saturn.services.views;
+
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.List;
+
+public class ViewRowCollection {
+
+    private Hashtable<String, List<ViewRow>> data = new Hashtable<String, List<ViewRow>>();
```
tgreenwood: Hashtable is a legacy class and not recommended for new code due to its synchronization overhead and other limitations, I would suggest HashMap here

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewRowCollection.java
tgreenwood: `data.computeIfAbsent(id, v ->  new ArrayList<>());` would be even better not to crate an instance of array list if not needed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewRowCollection.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public void add(String id, ViewRow row) {
+        if (!data.containsKey(id)) {
+            data.put(id, new ArrayList<>());
+        }
+        data.get(id).add(row);
+    }
+
+    public List<ViewRow> getRowsForId(String id) {
+        if (!data.containsKey(id)) {
```
tgreenwood: the same here, I would use `data.computeIfAbsent(id, v ->  new ArrayList<>());`

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -396,22 +397,17 @@ List<Map<String, Set<ValueDTO>>> retrieveJoinTableRows(
...
-        query.setString(1, id);
+        var query = getQuery(view, joinView, joinedTable, joinTable, projectionColumns, ids);
         var result = query.executeQuery();
-        List<Map<String, Set<ValueDTO>>> rows = new ArrayList<>();
+
+        var rows = new ViewRowCollection();
+
         while (result.next()) {
-            Map<String, Set<ValueDTO>> row = new HashMap<>();
+            var row = new ViewRow();
```
tgreenwood: for the sake of readability I would extract the row building to a separate method

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -445,10 +441,28 @@ List<Map<String, Set<ValueDTO>>> retrieveJoinTableRows(
         return rows;
     }
 
+    private PreparedStatement getQuery(String view, View.JoinView joinView, Table joinedTable, Table joinTable,
```
tgreenwood: hm-m, out of context the method name is not clear, I would add 'join' to the naming. What do you think?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -445,10 +441,28 @@ List<Map<String, Set<ValueDTO>>> retrieveJoinTableRows(
...
+            List<String> projectionColumns, List<String> ids) throws SQLException {
+
+
+        Function<String, String> queryBody = id
+                -> "(select " + projectionColumns.stream().map(String::toLowerCase).collect(Collectors.joining(", ")) + " , '" + id + "' AS rowid"
+                + " from " + joinedTable.name + " j "
+                + " where exists ("
```
tgreenwood, line 464: StringBuilder is more performant on many string appendings and requres less memory as '+' produce a new string due to immutable nature of string. Sometimes, compiler optimaize it and use stringbuilder under the hood, but not sure if it's been used here. I would explicitly use StringBuilder.

On the other hand, StringBuilder is less concise, but still I think it makes sense to go with it

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -539,6 +552,20 @@ public List<Map<String, Set<ValueDTO>>> retrieveRows(
...
     }
 
+    private void addJoinTableRowsForRow(
+            Map<String, Set<ValueDTO>> row,
+            List<ViewRow> allJoinTableRows) {
+
+        for (var joinTableRow : allJoinTableRows) {
+            joinTableRow.getRawData().forEach((key, values) -> {
+                if (!row.containsKey(key)) {
+                    row.put(key, new LinkedHashSet<>());
```
tgreenwood: computeIf absent

tgreenwood: not sure why linked hash set, do we need to keep order here?

PR #1470, opened 2024-01-22
Add support for external metadata views.

ewelinajozwik:
The changes are ready to be reviewed and tested.
Remaining tasks I am still working on:

- [x] Document the new feature
- [x] Add unit tests in Mercury


README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
     url: https://exstorage2/api/webdav
 ----
 
+=== Multi-Fairspace metadata views integration
+
+Fairspace can have only single data model configured. However, if there is a need to have data from multiple models,
```
frankyhollywood:
```suggestion
Fairspace works with a single data model. However, if there is a need to have data from multiple models,
```

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
...
+representing multiple domains that would be combined combined in single user interface, it is possible to integrate multiple Fairspace instances together.
+
+
+Fairspace (main) instance can be configured as a view interface of metadata from another (external) Fairspace instance.
```
frankyhollywood:
```suggestion
The main Fairspace instance can be configured as a view interface of metadata from another (external) Fairspace instance.
```

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
...
+External Fairspace can have a different metadata model configured than the main Fairspace.
+Important precondition is that the external instance has to be connected to the same Keycloak realm as the main instance.
```
frankyhollywood:
```suggestion
An important precondition is that the external instance has to be connected to the same Keycloak realm as the main instance.
```

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
...
+separate from internal metadata views, that allows to explore external metadata.
+Currently, cross-instance metadata search is not supported.
+
+
+For the external metadata pages, the tables and columns are created basing on the views configuration specified in that instance configuration.
+
+
+==== Access policy
+
+Connection to the same Keycloak realm allows to authenticate a user in all integrated Fairspace instances with a single set of login credentials.
```
frankyhollywood:
```suggestion
Connection to the same Keycloak realm allows to authenticate a user in all integrated Fairspace instances with a single set of login credentials (single sign-on).
```

projects/mercury/src/routes/WorkspaceRoutes.js
```diff
@@ -107,7 +119,7 @@ const WorkspaceRoutes = () => {
                 render={() => (isAdmin(currentUser) && (<UserRolesPage />))}
             />
 
-            <Redirect to="/workspaces" />
+            <Redirect to="/metadata-views" />
```
frankyhollywood: is this correct?

frankyhollywood (approved):
Looks good, haven't tested it though but we will deploy soon on dev environment

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
...
+| ``label``
+| String to be used as a display name of the metadata source.
+| ``url``
+| Fairspace instance to connect to.
+|===
+
+Sample configuration of external metadata sources in YAML format:
+[source, bash]
+----
+metadata-sources:
```
frankyhollywood, line 1879: So this is in values.yaml? And do you also need to add the local instance, or only remotes?

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
     url: https://exstorage2/api/webdav
 ----
 
+=== Multi-Fairspace metadata views integration
+
+Fairspace can have only single data model configured. However, if there is a need to have data from multiple models,
+representing multiple domains that would be combined combined in single user interface, it is possible to integrate multiple Fairspace instances together.
```
tgreenwood: "combined combined" - one is redundant

README.adoc
```diff
@@ -1833,6 +1833,60 @@ storages:
...
+
+
+Fairspace (main) instance can be configured as a view interface of metadata from another (external) Fairspace instance.
+External Fairspace can have a different metadata model configured than the main Fairspace.
+Important precondition is that the external instance has to be connected to the same Keycloak realm as the main instance.
+
+
+External metadata can be searched and browsed in the same way as the internal metadata views. When configured, there is an extra page,
```
tgreenwood, line 1847: do you think it makes sense to add a picture of the interface for a few models here or at the beginning of the doc?

projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
```diff
@@ -29,6 +37,7 @@ public static class Storage {
     private boolean forceHttps = true;
     private List<String> domains = Collections.emptyList();
     private Map<String, Storage> storages = new LinkedHashMap<>();
+    private Map<String, MetadataSource> metadataSources = new LinkedHashMap<>();
```
tgreenwood, line 40: why do you use linked map?

projects/pluto/src/main/java/nl/fairspace/pluto/web/MetadataSourcesResource.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.stream.Collectors;
+
+import static nl.fairspace.pluto.config.Urls.METADATA_SOURCES_PATH;
+
+@RestController
+@Slf4j
+public class MetadataSourcesResource {
+
+    @Autowired
+    PlutoConfig plutoConfig;
```
tgreenwood: Can be private I think. Also, no need to put annotation over it, just define a constructor with @AllArgsContructor - spring automatically autowires

projects/pluto/src/main/java/nl/fairspace/pluto/web/dto/MetadataSourceInfo.java
```diff
@@ -0,0 +1,16 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+@Getter
+@EqualsAndHashCode
+@AllArgsConstructor
+@NoArgsConstructor
```
tgreenwood: do we need no args constructor if there is no setter?

tgreenwood (commented):
Looks good, just a few minor comments.

projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
ewelinajozwik: to maintain the order of configured sources, but also to be consistent with other configuration options here (of same type).

projects/mercury/src/routes/WorkspaceRoutes.js
ewelinajozwik: oups, no, included by mistake, I'll revert it.

README.adoc
ewelinajozwik: only remotes 
I forgot to add it to Kubernetes/Helm config, will do!

README.adoc
ewelinajozwik: Not sure... we don't have a proper demo instance to take the screenshot of...

projects/pluto/src/main/java/nl/fairspace/pluto/web/MetadataSourcesResource.java
ewelinajozwik: Are you sure? I haven't tried that yet... Also we do it his way in all other places, so it would be good to change them all then for consistency

projects/pluto/src/main/java/nl/fairspace/pluto/web/dto/MetadataSourceInfo.java
ewelinajozwik: I'll remove it

projects/pluto/src/main/java/nl/fairspace/pluto/web/MetadataSourcesResource.java
ewelinajozwik: Also, this is not a standard class, but RestController, there is no standard constructor for it

projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
tgreenwood: I think that consistency should not be an argument at choosing a data structure, but minimal complexity of operations over the data structure for the particular case. But good if order does matter here

projects/pluto/src/main/java/nl/fairspace/pluto/web/MetadataSourcesResource.java
tgreenwood: I am sure :) this is a relatively old spring's feature. Also, it is recommended to put the autowired annotation over either constructor or setter as, otherwise, for (at least) private fields it will be using reflection, what slows down the init stage.

Regarding 'standard constructor': from Spring's perspective it is just a regular class, Spring uses constructor you provide for bean creation. Then populate fields (if not yet within constructor with parameters). In this case it will be using no args constructor and then (not sure) set it directly (as it is of default access modifier) or within reflection

tgreenwood: So, it should work if we put it like this:

```
@RequiredArgsConstructor
public class MetadataSourcesResource {

    private PlutoConfig plutoConfig;
    ....
}
```
Spring will resolve on its own

ewelinajozwik: this gives me

> java.lang.NullPointerException: Cannot invoke "nl.fairspace.pluto.config.dto.PlutoConfig.getMetadataSources()" because "this.plutoConfig" is null
        at nl.fairspace.pluto.web.MetadataSourcesResource.metadataSources(MetadataSourcesResource.java:32) ~[main/:na]
        Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:

`@AllArgsConstructor` annotation seems to work fine, but that might be problematic, when adding new private fields to this class. I can add it, but to me it is less clear then what is being autowired here...

ewelinajozwik: but you're right @tgreenwood, constructor injection is indeed a recommended way...
I'll just put the constructor there instead of the annotation:
```
    private final PlutoConfig plutoConfig;
  
    public MetadataSourcesResource(PlutoConfig plutoConfig) {
        this.plutoConfig = plutoConfig;
    }
```

tgreenwood: Sorry, I missed `final` in my piece of code, in this case it would work you with `@RequiredArgsContructor`. And also all fields are supposed to be final too, then it is not a problem

tgreenwood: but feel free to leave it the way you have put it, not to stuck on this not critical comment

tgreenwood (approved):

PR #1472, opened 2024-01-23
FAIRPC-29: added tests for caching, made cached configurable and slig…

tgreenwood:
…htly refactored

### DIFF
--- a/projects/saturn/application.yaml
+++ b/projects/saturn/application.yaml
@@ -61,9 +61,7 @@ search:
   countRequestTimeout: 60000
 caches:
   facets:
-    enabled: true
     name: "facets"
   views:
-    enabled: true
     name: "views"
 
--- a/projects/saturn/application.yaml
+++ b/projects/saturn/application.yaml
@@ -59,3 +59,11 @@ viewDatabase:
 search:
   pageRequestTimeout: 10000
   countRequestTimeout: 60000
+caches:
+  facets:
+    enabled: true
+    name: "facets"
+  views:
+    enabled: true
+    name: "views"
+
--- a/projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
@@ -13,6 +13,9 @@ import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
 import org.apache.jena.atlas.json.JSON;
 import org.apache.jena.tdb2.params.StoreParams;
 import org.apache.jena.tdb2.params.StoreParamsCodec;
@@ -49,6 +52,8 @@ public class Config {
     @JsonSetter(nulls = Nulls.AS_EMPTY)
     public Map<String, String> services = new HashMap<>();
 
+    public Caches caches = new Caches();
+
     public Search search = new Search();
 
     public static class Jena {
@@ -75,6 +80,20 @@ public class Config {
 
     public static class WebDAV {
         public String blobStorePath = "data/blobs";
+    }
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class CacheConfig {
+        public String name;
+        public boolean enabled;
+        public boolean autoRefreshEnabled;
+        public Long refreshFrequencyInHours;
+    }
+
+    public static class Caches {
+        public CacheConfig facets = CacheConfig.builder().name("facets").enabled(true).build();
+        public CacheConfig views = CacheConfig.builder().name("views").enabled(true).build();
     }
 
     public static class Search {
--- a/projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
@@ -22,7 +22,11 @@ import org.apache.jena.tdb2.params.StoreParamsCodec;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.*;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class Config {
     static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory())
@@ -86,14 +90,13 @@ public class Config {
     @AllArgsConstructor
     public static class CacheConfig {
         public String name;
-        public boolean enabled;
-        public boolean autoRefreshEnabled;
-        public Long refreshFrequencyInHours;
+        public boolean autoRefreshEnabled = false;
+        public Long refreshFrequencyInHours = 240L;
     }
 
     public static class Caches {
-        public CacheConfig facets = CacheConfig.builder().name("facets").enabled(true).build();
-        public CacheConfig views = CacheConfig.builder().name("views").enabled(true).build();
+        public CacheConfig facets = CacheConfig.builder().name("facets").build();
+        public CacheConfig views = CacheConfig.builder().name("views").build();
     }
 
     public static class Search {
--- a/projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
@@ -24,6 +24,8 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 import static io.fairspace.saturn.config.ViewsConfig.ColumnType;
 import static io.fairspace.saturn.config.ViewsConfig.View;
@@ -52,16 +54,16 @@ public class ViewService {
             """, FS.NS));
 
     private static final Query RESOURCE_TYPE_VALUES_QUERY = QueryFactory.create(String.format("""
-          PREFIX fs: <%s>
-          SELECT ?value ?label
-          WHERE {
-             VALUES (?value ?label) {
-                (fs:Collection "Collection")
-                (fs:Directory "Directory")
-                (fs:File "File")
-             }
-          }
-          """, FS.NS));
+            PREFIX fs: <%s>
+            SELECT ?value ?label
+            WHERE {
+               VALUES (?value ?label) {
+                  (fs:Collection "Collection")
+                  (fs:Directory "Directory")
+                  (fs:File "File")
+               }
+            }
+            """, FS.NS));
 
     private static final Query BOUNDS_QUERY = QueryFactory.create(String.format("""
             PREFIX fs: <%s>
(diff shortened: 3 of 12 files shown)


projects/saturn/application.yaml
```diff
@@ -59,3 +59,11 @@ viewDatabase:
 search:
   pageRequestTimeout: 10000
   countRequestTimeout: 60000
+caches:
```
frankyhollywood, line 62: I wonder if this should be a feature toggle, it is extra code and configuration while I assume it will always be activated.

What do you think @ewelinagr ?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
...
+            viewsCache.refresh(Boolean.TRUE);
+        }
         FilteredDatasetGraph.enableQuadPermissionCheck();
         log.info("Caches refreshing/warming up successfully finished");
     }
 
-    public List<FacetDTO> getCachedFacets() {
-        try {
+    public List<FacetDTO> getFacets() {
+        if (cachesConfig.facets.enabled) {
```
frankyhollywood: what happens if cache is disable, no permission check is performed?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
...
+                // this check is needed for cached data only as, otherwise,
+                // the check will be performed during retrieving data from Jena
+                throw new AccessDeniedException(USER_DOES_NOT_HAVE_PERMISSIONS_TO_READ_FACETS);
+            }
+            try {
+                return facetsCache.get(Boolean.TRUE);
+            } catch (ExecutionException e) {
+                throw new RuntimeException(e);
             }
-            return facetsCache.get(Boolean.TRUE);
```
frankyhollywood, line 122: do you know what this old code is, was there already a facet cache? And if so, why did we need a new cache?

frankyhollywood:
overal looks good! A few comments see above

frankyhollywood:
Another question I have is how cache is invalidated, sometimes a pipeline adds new data to Fairspace, how does the cache respond to this?

frankyhollywood:
impressive amount of tests btw! :)

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
                FILTER NOT EXISTS { ?subject fs:dateDeleted ?anyDateDeleted }
             }
             """, FS.NS));
+    public static final String USER_DOES_NOT_HAVE_PERMISSIONS_TO_READ_FACETS = "User does not have permissions to read facets";
```
frankyhollywood, line 86: personally for me this adds little value, unless there is a global exception message provider, so through the whole application we use the same messages

projects/saturn/application.yaml
tgreenwood: Well, to be honest I also think that it will be enabled permanently as we definitely won't have memory problems for this amount of cached data. And yes, we have to add a bunch of if's in the code to support the toggling functionality. Personally, I would get rid of toggle but left configuration as it might be extended one day in the future (I don't believe it, but this is still a good practice :) )

@ewelinagr could you please share your vision?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: not really, permission check is performed in FilteredDatasetGraph - for each particular quad retrieved from Jena

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: No-no, it is new one, the first iteration which's been merged just yesterday. In the second iteration I slightly refactored it

tgreenwood:
> Another question I have is how cache is invalidated, sometimes a pipeline adds new data to Fairspace, how does the cache respond to this?

There are three ways to refresh the cache:
- call the reindex endpoint
- restart Fairspace as cache is warmed up on startup
- configure auto-refreshing including how often you want it to be refreshed, then it will be refreshed automatically

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: I use the const in test to check the message of thrown exception

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
frankyhollywood: I don't see it, not in 'fetchFacets' or in the 'calculateRead', where is it supposed to be checked?

tgreenwood: check `FilteredDatasetGraph` class, it is a filter on any read from Jena

projects/saturn/application.yaml
ewelinajozwik: To me it is also fine to keep the caching permanently enabled, the only comment I had around config for it was to make the time of automatic refresh of caches configurable (in the first draft it was hardcoded and set to daily refresh) and I still think this auto-refresh setup part should not be fixed

tgreenwood: Then are we okay if I remove the enabled flag and make caching permanently available but leave the part about auto-refresh configurable and switched off be default? @FrankyHollywood  @ewelinagr

ewelinajozwik: I agree. Also, I don't think it has to be done per cache type (separate per facets, separate per views), can be single config for all the caches

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
...
+            this.facetsCache = buildCache(this::fetchFacets, cachesConfig.facets);
+        }
+        if (cachesConfig.views.enabled) {
+            this.viewsCache = buildCache(this::fetchViews, cachesConfig.views);
+        }
         refreshCaches();
     }
 
     public void refreshCaches() {
         log.info("Caches refreshing/warming up has been triggered");
```
ewelinajozwik, line 112: This log is shown even if there are no caches enabled.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
...
         FilteredDatasetGraph.disableQuadPermissionCheck();
-        facetsCache.refresh(Boolean.TRUE);
-        viewsCache.refresh(Boolean.TRUE);
+        if (cachesConfig.facets.enabled) {
+            facetsCache.refresh(Boolean.TRUE);
+        }
+        if (cachesConfig.views.enabled) {
+            viewsCache.refresh(Boolean.TRUE);
+        }
         FilteredDatasetGraph.enableQuadPermissionCheck();
```
ewelinajozwik: not sure if there is any error possible during cache refreshing, but to make sure the `enableQuadPermissionCheck` will be triggered no matter what, maybe it might be better to wrap the refreshing in try/catch and do `enableQuadPermissionCheck` in `finally`?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -81,55 +83,76 @@ public class ViewService {
...
         this.metadataPermissions = metadataPermissions;
-        this.facetsCache = buildFacetCache();
-        this.viewsCache = buildViewsCache();
+        if (cachesConfig.facets.enabled) {
+            this.facetsCache = buildCache(this::fetchFacets, cachesConfig.facets);
+        }
+        if (cachesConfig.views.enabled) {
+            this.viewsCache = buildCache(this::fetchViews, cachesConfig.views);
+        }
         refreshCaches();
```
ewelinajozwik, line 108: I am a bit confused about the order of operations in this init.
The caches are first build, then refresh on startup, right?
In my logs I see:

> 2024-01-24 11:57:02 INFO  Caches refreshing/warming up has been triggered
2024-01-24 11:57:02 INFO  List of facets has been cached, 1 facets in total
2024-01-24 11:57:02 INFO  List of views has been cached, 4 views in total
2024-01-24 11:57:02 INFO  Caches refreshing/warming up successfully finished

which is a mix of the two operations: first start refreshing -> then build -> then finish refreshing

Is this for sure correct?

projects/saturn/src/test/java/io/fairspace/saturn/services/metadata/MetadataPermissionsTest.java
```diff
@@ -0,0 +1,360 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MetadataPermissionsTest {
```
ewelinajozwik, line 25: Really nice test coverage!
It looks like the tests around workspaces and dav resources access might be duplicating part of DavFactoryAccessTest, but we can consider these more unit tests.

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
```diff
@@ -76,6 +81,20 @@ public static class Auth {
     public static class WebDAV {
         public String blobStorePath = "data/blobs";
     }
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class CacheConfig {
+        public String name;
+        public boolean enabled;
+        public boolean autoRefreshEnabled;
```
ewelinajozwik: please set reasonable defaults here, since application.yaml gets fully overwritten when deploying with helm.

ewelinajozwik: It would be good to include these in helm deployment setup too

ewelinajozwik:
@tgreenwood I'm done reviewing, please see my comments above

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: very good catch, super makes sense, added

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: `this::fetchViews` does not call the method right away, it is just a reference, so the build method just keeps the ref. the fetch method is called on refresh

tgreenwood: please contact me if still not clear, resolving not to be blocked

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
tgreenwood: Ddefaults value added, but did not include them in helm since not sure I understood correctly the comment. Resolving not to be blocked.

projects/saturn/src/test/java/io/fairspace/saturn/services/metadata/MetadataPermissionsTest.java
tgreenwood: thanks! this is truly unit tests

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: since now cache is enabled permanently (as agreed)

projects/saturn/application.yaml
tgreenwood: since now caches are enabled permanently with no toggle

frankyhollywood (approved):

PR #1473, opened 2024-01-23
New indexes for jointables

frankyhollywood:


projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -103,62 +102,103 @@ void ensureTableExists(Table table) throws SQLException {
...
-                    connection.createStatement().execute(command);
-                    connection.setAutoCommit(false);
-                    log.info("Table {} updated.", table.name);
-                }
+                updateTable(table, connection);
             }
         }
     }
 
+    void ensureJoinTableExists(Table table) throws SQLException {
```
tgreenwood: I would call this method something like createOrUpdateTable

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -103,62 +102,103 @@ void ensureTableExists(Table table) throws SQLException {
...
+
+    private void updateTable(Table table, Connection connection) throws SQLException {
+        var columnMetadata = getColumnMetadata(connection, table.name);
+        var newColumns = table.columns.stream()
+                .filter(column -> !column.type.isSet() && !columnMetadata.containsKey(column.name))
+                .collect(Collectors.toList());
+        try {
+            log.debug("New columns: {}", new ObjectMapper().writeValueAsString(newColumns));
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
tgreenwood: could you please replace it using log.error? I saw it many times, but once you are touching this part

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -103,62 +102,103 @@ void ensureTableExists(Table table) throws SQLException {
...
+        
+        var keys = table.columns.stream()
+                .filter(column -> column.type == ColumnType.Identifier)
+                .map(column -> column.name)
+                .toList();
+
+        connection.setAutoCommit(true);
+
+        for (var column : keys) {
+            var indexName = String.format("%1$s_%2$s_idx", table.name, column);
```
tgreenwood: `%1$s_%2$s_idx` could be replaced with  just`%s_%s_idx`

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -103,62 +102,103 @@ void ensureTableExists(Table table) throws SQLException {
...
+            log.info("Index {} created.", indexName);
+        }
+        
+        connection.setAutoCommit(false);
+    }
+
     void validateViewConfig(ViewsConfig.View view) {
-        if (view.columns.stream().anyMatch(column ->
-                column.name.equalsIgnoreCase("id"))) {
+        if (view.columns.stream().anyMatch(column -> column.name.equalsIgnoreCase("id"))) {
```
tgreenwood: please, extract variable with proper naming for batter readability

tgreenwood: and in other if conditions below

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -103,62 +102,103 @@ void ensureTableExists(Table table) throws SQLException {
...
             throw new IllegalArgumentException(
                     "Forbidden to override the built-in column 'id' of view " + view.name);
         }
-        if (view.columns.stream().anyMatch(column ->
-                column.name.equalsIgnoreCase("label"))) {
+        if (view.columns.stream().anyMatch(column -> column.name.equalsIgnoreCase("label"))) {
```
tgreenwood: good practice is to put const on the left side like `"label".equalsIgnoreCase(column.name)` what helps to avoid NPE. Maybe here NPE is not the case, but this is just a common rule

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -224,14 +264,14 @@ public static Table getJoinTable(View.JoinView join, View view) {
...
                 .build();
 
         return joinTable;
     }
 
-    @Data @Builder
+    @Data
+    @Builder
     public static class ColumnMetadata {
         String type;
```
tgreenwood: let's make fields private then

tgreenwood (commented):
looks good, just a few minor comments

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
FrankyHollywood: "I saw it many times", you mean in general or Fairspace code? I think it's rare but I will change!

frankyhollywood: its an exception if a 'log.debug' doesn't work out, that should not be reported as an error

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
frankyhollywood: yes that's right, I'll change

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
frankyhollywood: for complex cases I would agree, however in this case it is just a single line with a single string comparison, I think any developer will understand and will not improve readability since you would add 4 lines of code just for 4 string comparisons.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
frankyhollywood: good one, I'll change

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
frankyhollywood: as discussed, let's log an error and rethrow, because it should never fail indeed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
frankyhollywood: good suggestion, will change!

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -97,68 +96,111 @@ Map<String, ColumnMetadata> getColumnMetadata(Connection connection, String tabl
...
+        var keys = table.columns.stream()
+                .filter(column -> column.type == ColumnType.Identifier)
+                .map(column -> column.name)
+                .toList();
+
+        connection.setAutoCommit(true);
+
+        for (var column : keys) {
+            var indexName = String.format("%s_%s_idx", table.name, column);
+            var command = String.format("CREATE INDEX IF NOT EXISTS %s ON public.%s (%s)", indexName, table.name, column);
```
tgreenwood: schema name "public" is redundant, and it even can be different on env

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClientFactory.java
```diff
@@ -224,15 +266,15 @@ public static Table getJoinTable(View.JoinView join, View view) {
...
 
         return joinTable;
     }
 
-    @Data @Builder
-    public static class ColumnMetadata {
+    @Data
+    @Builder
+    private static class ColumnMetadata {
         String type;
```
tgreenwood: forgot to make fields private

PR #1481, opened 2024-02-05
Performance improvements

tgreenwood:
Introduced two more materialized views to speed up metadata views loading


frankyhollywood (approved):

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
```diff
@@ -102,6 +102,7 @@ public static class Caches {
     public static class Search {
         public long pageRequestTimeout = 10_000;
         public long countRequestTimeout = 100_1000;
+        public int maxJoinItems = 50; // used to limit number of joined entries (from the join view) to decrease the response size
```
ewelinajozwik, line 105: Could you document in README too? 
We have a sample config there with comments/descriptions of some options.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -426,63 +417,49 @@ private ViewRowCollection retrieveJoinTableRows(
...
-    private ViewRow buildJoinRows(View.JoinView joinView, Set<String> projectionColumns, ResultSet result) throws SQLException {
+    private ViewRow buildJoinRows(View.JoinView joinView, String joinViewId, Set<String> projectionColumns, ResultSet result) throws SQLException {
 
         var row = new ViewRow();
-        row.put(joinView.view, Sets.newHashSet(new ValueDTO(result.getString("label"), result.getString("id"))));
+        row.put(joinView.view, Sets.newHashSet(new ValueDTO(result.getString(joinView.view + "_label"), result.getString(joinViewId))));
 
         for (var column : projectionColumns) {
-            var columnName = joinView.view + "_" + column;
+//            var columnName = joinView.view + "_" + column;
```
ewelinajozwik: to be removed?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
```diff
@@ -554,7 +531,7 @@ public List<Map<String, Set<ValueDTO>>> retrieveRows(String view, List<ViewFilte
                     .map(ViewRow::getRawData)
                     .toList();
         } catch (SQLException e) {
-            throw new QueryException("Error retrieving page rows", e);
+            throw new RuntimeException("Error retrieving page rows", e);
```
ewelinajozwik: Why not QueryException?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewUpdater.java
```diff
@@ -45,6 +59,7 @@ public void close() throws SQLException {
 
     public void commit() throws SQLException {
         viewStoreClient.commit();
+        materializedViewService.createOrUpdateAllMaterializedViews();
```
ewelinajozwik: why here, not witthin `viewStoreClient.commit();` function? There are some conditions on when to update, e.g. check on transaction mode or if is extrastoragetransaction. I think the condition on when to update materializedViews should be the same as for views...

projects/saturn/src/main/java/io/fairspace/saturn/services/metadata/MetadataApp.java
```diff
@@ -4,22 +4,33 @@
...
-import static spark.Spark.*;
+import static spark.Spark.delete;
+import static spark.Spark.get;
+import static spark.Spark.patch;
+import static spark.Spark.put;
 
 @Log4j2
 public class MetadataApp extends BaseApp {
+
+    private static final String DO_VIEWS_UPDATE = "doViewsUpdate";
```
ewelinajozwik, line 32: Please document it in our documentation.

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
frankyhollywood: I was actually wondering why this needs to be configurable, just add complexity, while it is not likely to ever change.

tgreenwood: I missed the part, it should be also in the materialized view generation instead of hardcoded number (as it is now). I think it makes sense to have it configurable as we create join materialized views with number of data limited to the number defined in the config. For less resources we would be able even to decrease the number of joins

tgreenwood: I will add it to readme

tgreenwood: @ewelinagr I do not see application.yaml example in our readme. What did you mean? values.yaml?

ewelinajozwik: Ok, I thought we had it... Not sure where to document it. 
No, values.yaml is for helm/kubernetes setup, but you didn't make it helm-configurable, so it is not a right place.

tgreenwood: Added the parameter to materialzed view building too.

@ewelinagr let's then leave it as internal parameter. Otherwise, it would make sense to describe all the others too. What do you think?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
tgreenwood: yep, thanks, removed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreReader.java
tgreenwood: it was replaced for profiling needs and missed when I was rolling it back. Switched back to QueryException

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewUpdater.java
tgreenwood: because materialized views need the data to be committed to be build on top of the data

tgreenwood: Screenshot 2024-02-13 at 14 35 18
The condition is check a call upstream, do you see any others checks to be performed?

projects/saturn/src/main/java/io/fairspace/saturn/services/metadata/MetadataApp.java
tgreenwood: good catch, done

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewUpdater.java
ewelinajozwik: I meant those 2 conditions. Updating should not happen if it's and extra storage transaction and if not write transaction.
Not sure if I get what it means that it needs data...
On the commit level you also have the data, including list of subjects

tgreenwood: @ewelinagr agree, it makes much more sense to call the mv update on commit of the ViewStoreClient.

Regarding conditions, I added a screen above showing that it is already there

ewelinajozwik: ok, I get it now, I misread the order of commit() function calls

PR #1482, opened 2024-02-06
Improve external metadata view error handling.

ewelinajozwik:


projects/mercury/src/common/utils/httpUtils.js
```diff
@@ -51,6 +51,25 @@ export function handleHttpError(defaultMessage) {
     };
 }
 
+export function handleRemoteSourceHttpError(defaultMessage) {
```
tgreenwood, line 54: It would be useful to have a comment over the method explaining why it is different for external source

PR #1483, opened 2024-02-12
Dashboard first version

frankyhollywood:


ewelinajozwik:
I don't like the version being moved to the user details panel like this, at least not with the current formatting:
image

It looks like additional user being logged in...
If you want to have it there, instead of the footer, I would put it below the logout button. And make it look like info text, not like clickable menu option looking the same as user name...

projects/mercury/src/dashboard/DashboardPage.js
```diff
@@ -0,0 +1,76 @@
...
+import {APPLICATION_NAME, METADATA_VIEW_MENU_LABEL} from "../constants";
+
+const DashboardPage = (props) => {
+    const {currentUser, classes} = props;
+    const {views} = useContext(MetadataViewContext);
+    const {externalMetadataSources} = useContext(ExternalMetadataSourceContext);
+    const canViewMetadata = currentUser && currentUser.canViewPublicMetadata && views && views.length > 0;
+
+    return (
+        <Grid container justifyContent="center" spacing="5">
```
ewelinajozwik, line 20: this whole content is visible even if not logged in (or backend is down).
Not sure if this is a good idea. I would distinguish those states and if backend is down or not logged in, show something else (e.g. info to log in to see the whole content or something like that).

projects/mercury/src/dashboard/DashboardPage.js
```diff
@@ -0,0 +1,76 @@
...
+            <BreadCrumbs />
+            <Paper className={classes.mainPage}>
+                <Grid container justifyContent="center" spacing="5">
```
ewelinajozwik, line 23: Will you move it left as discussed?

projects/mercury/src/dashboard/DashboardPage.js
```diff
@@ -0,0 +1,76 @@
...
+                        {canViewMetadata && (
+                            <DomainInfo
+                                domainName={METADATA_VIEW_MENU_LABEL}
+                                domainLink="/metadata-views"
+                            />
+                        )}
+                        {canViewMetadata && externalMetadataSources.map((source) => (
+                            <DomainInfo
+                                domainName={source.label}
+                                domainLink={source.name}
```
ewelinajozwik: I don't think this is the proper link.
It is not working.

projects/mercury/src/dashboard/DashboardPage.js
frankyhollywood: In slack Elisa mentioned she liked centered view. Let's leave it for now. Next week we will have a more in depth design discussion with Elisa and Elena.

projects/mercury/src/dashboard/DashboardPage.js
frankyhollywood: Have you tested? It's a relative link, and seems to work

ewelinajozwik: yes, I tested and it works only for internal metadata, not external one.
For external metadata it redirects to `/<source_name>`, while the path should be` /metadata-sources/<source_name>`

projects/mercury/src/layout/UserMenu.js
```diff
@@ -91,19 +99,31 @@ const UserMenu = ({classes}) => {
...
-                                    </MenuItem>
-                                    <Divider />
-                                    <MenuItem onClick={handleLogout}>
-                                        <ListItemIcon>
-                                            <ExitToAppIcon />
-                                        </ListItemIcon>
-                                        <ListItemText primary="Logout" />
+                                    <MenuItem
+                                        className={classes.userMenu}
+                                        disableFocusOnHover
```
ewelinajozwik: There is a js warning in the console because of this:

> Warning: React does not recognize the `disableFocusOnHover` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `disablefocusonhover` instead.

projects/mercury/src/dashboard/DashboardPage.js
```diff
@@ -0,0 +1,76 @@
...
+                </Grid>
+            </Paper>
+        </Grid>
+    );
+};
+
+const ContextualDashboardPage = (props) => {
+    const {currentUser} = useContext(UserContext);
+
+    return <DashboardPage currentUser={currentUser} {...props} />;
```
ewelinajozwik, line 73: not sure why, but somewhere here there is a js warning:
> Warning: Each child in a list should have a unique "key" prop.
Check the render method of `DashboardPage`. See https://reactjs.org/link/warning-keys for more information.
    at WithStyles (http://localhost:3000/static/js/bundle.js:86329:21)
    at DashboardPage (http://localhost:3000/static/js/bundle.js:9466:27)
    at ContextualDashboardPage (http://localhost:3000/static/js/bundle.js:9635:71)
    at WithStyles (http://localhost:3000/static/js/bundle.js:86329:21)
    at Route (http://localhost:3000/static/js/bundle.js:217149:29)
    at Switch (http://localhost:3000/static/js/bundle.js:217351:29)
    at WorkspaceRoutes (http://localhost:3000/static/js/bundle.js:36336:70)

projects/mercury/src/layout/UserMenu.js
frankyhollywood: good catch, will change

projects/mercury/src/dashboard/DashboardPage.js
frankyhollywood: Not sure how to solve, let's discuss

ewelinajozwik (approved):

PR #1486, opened 2024-02-16
FAIRPC-29: fixed views and mv's update

tgreenwood:


projects/saturn/src/main/resources/log4j2.properties
```diff
@@ -18,7 +18,7 @@ appender.stdout.type = Console
 appender.stdout.name = stdout
 appender.stdout.target = System.out
 appender.stdout.layout.type = PatternLayout
-appender.stdout.layout.pattern = %d{yyyy-MM-dd HH:mm:ss} %-5p %m%n
```
ewelinajozwik, line 21: what does %t do?

tgreenwood: just add current thread name to the log record, needed for systems with more that one thread

ewelinajozwik (approved):

PR #1493, opened 2024-02-27
Added spotless as a code format validator and auto-formatter to Saturn for Java

tgreenwood:


projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
```diff
@@ -102,10 +103,10 @@ public static class Caches {
     public static class Search {
         public long pageRequestTimeout = 10_000;
         public long countRequestTimeout = 100_1000;
-        public int maxJoinItems = 50; // used to limit number of joined entries (from the join view) to decrease the response size
+        public int maxJoinItems =
+                50; // used to limit number of joined entries (from the join view) to decrease the response size
```
ewelinajozwik: here moving 50 to the next line feels strange to me

ewelinajozwik (approved):
Other than the odd formatting suggestion that I mentioned in the comment it looks fine to me.

projects/saturn/src/main/java/io/fairspace/saturn/config/Config.java
tgreenwood: Totally agree. Fixed it manually, there is no such a granular toggle to configure formatting. It won't be a big issue once applied the formatting once. Then such cases will be easily caught on code review

PR #1494, opened 2024-02-27
Added spotless as a code format validator and auto-formatter for Pluto

tgreenwood:


projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
```diff
@@ -1,34 +1,45 @@
...
 @Configuration
 @Data
 public class PlutoConfig {
     @Data
     @JsonInclude(JsonInclude.Include.NON_NULL)
     public static class MetadataSource {
-        @NotBlank private String name;
-        @NotBlank private String label;
-        @NotBlank private String url;
+        @NotBlank
```
ewelinajozwik, line 21: @tgreenwood regarding type annotations - indeed it looks better in this case, if they are inline, but I'm ok with having them in separate lines too. 
At least it looks like when such annotations are used in constructors, it is allowed to keep them inline (e.g. https://github.com/thehyve/fairspace/pull/1493/files#diff-6a38f9f2003282b1b73e5b098546d6cd2c879c3bc5890755ffcd88639b623f09R36)

ewelinajozwik (approved):

PR #1500, opened 2024-03-12
tuned ESlint config (mostly removed work-arounds) and fixed code with…

tgreenwood:
… the updated config


ewelinajozwik (commented):
Maybe it's my personal feeling, but I'm not convinced to those changes.
I really don't like unnecessary spacing, also enforcing a space after `{ ` feels inconsistent to me - there is no empty space after `[` or `(`.
Also putting a comma at the end of the list seems unnecessary...
Is it really enforced convention of react/js? 

Why you called the previous settings "work-arounds"? "object-curly-spacing" = never is one of regular ESLint config options

tgreenwood:
> Maybe it's my personal feeling, but I'm not convinced to those changes. I really don't like unnecessary spacing, also enforcing a space after `{ ` feels inconsistent to me - there is no empty space after `[` or `(`. Also putting a comma at the end of the list seems unnecessary... Is it really enforced convention of react/js?
> 
> Why you called the previous settings "work-arounds"? "object-curly-spacing" = never is one of regular ESLint config options

@ewelinagr thanks for feedback. Agree, it is not fair to generalize and call all of them 'work-arounds', my bad, just some of them like quotes, max-len (which I did not change as it will be fixed automatically with prettier once introduced). 

Both of the mentioned rules come from the airbnb config. I am absolutely fine to rollback them as I agreed with 

- the spacing before and after curve brackets because of Java background (here is my personal feeling, but I accept that JS is another world), 
- and with the commas following arguments that it produces less changes lines when you need to add more pairs (I don't like how it looks like too, but was convinced with the argument behind)

Before I rolled back these two rules, how do you feel about others?

projects/mercury/src/common/utils/__tests__/genericUtils.js
```diff
@@ -87,7 +87,7 @@ describe('isNonEmptyValue', () => {
         values.forEach(v => expect(isNonEmptyValue(v)).toBe(true));
     });
     it('Returns false for the given values', () => {
-        const values = [undefined, null, '', NaN, "", ``];
+        const values = [undefined, null, '', NaN, '', ''];
```
frankyhollywood: the equivalent quotes can be removed, not sure what the original intent was, especially ``

frankyhollywood (commented):
eslint REALLY doesn't like double quotes 😂

frankyhollywood (approved):

projects/mercury/src/common/utils/__tests__/genericUtils.js
tgreenwood: good point, rolled back the line and disable eslint for the line

tgreenwood:
> eslint REALLY doesn't like double quotes 😂

Well, yes, but it is easy to disable the rule, but I thought it would be better to use the same quotes everywhere for consistency. Feel free to disagree

ewelinajozwik:
> > Maybe it's my personal feeling, but I'm not convinced to those changes. I really don't like unnecessary spacing, also enforcing a space after `{ ` feels inconsistent to me - there is no empty space after `[` or `(`. Also putting a comma at the end of the list seems unnecessary... Is it really enforced convention of react/js?
> > Why you called the previous settings "work-arounds"? "object-curly-spacing" = never is one of regular ESLint config options
> 
> @ewelinagr thanks for feedback. Agree, it is not fair to generalize and call all of them 'work-arounds', my bad, just some of them like quotes, max-len (which I did not change as it will be fixed automatically with prettier once introduced).
> 
> Both of the mentioned rules come from the airbnb config. I am absolutely fine to rollback them as I agreed with
> 
> * the spacing before and after curve brackets because of Java background (here is my personal feeling, but I accept that JS is another world),
> * and with the commas following arguments that it produces less changes lines when you need to add more pairs (I don't like how it looks like too, but was convinced with the argument behind)
> 
> Before I rolled back these two rules, how do you feel about others?

Ah ok, thanks for the clarification!
You might be right that a white space before/after `{` is java inspired, although in java you don't see that inline often, at least not as often as in js and to me it just looks odd...

Regarding commas - I get the point, but not so convincing to me - following this idea we could also allow to leave a dangling comma at the end of a list (if multi-line) in case we want to add something later (['a', 'b',])

I am fine with the changes that are now left! Although strange that double-quotes will be now forbidden :sweat_smile:

ewelinajozwik (approved):

PR #1502, opened 2024-03-13
Improvement/fairpc 38 prettier

tgreenwood:


projects/mercury/src/permissions/PermissionViewer.js
```diff
@@ -14,20 +14,26 @@ const useStyles = makeStyles({
...
-    collection, workspaceUsers, collaboratingWorkspaces,
-    collaboratingUsers, currentUser, setPermission, error, loading
+    collection,
+    workspaceUsers,
+    collaboratingWorkspaces,
+    collaboratingUsers,
+    currentUser,
+    setPermission,
+    error,
+    loading
```
ewelinajozwik, line 28: In general I'm ok with the changes in this PR, but they are really strict now and in places like this, for example, I prefer to have all the properties next to each other rather than separate lines. This takes a lot of space for each class and makes those less readable to me.

tgreenwood: this happens only when properties being placed next to each other they take more then 120 symbols. I set max line length to 120, so it's been formatted. So, in case of, let's say, just 3-4 properties taking less than 120 symbols they will be on the same like even if you put them on different ones

tgreenwood: unfortunately, the only way to change it is to increase the max length, but 120 is already a big number

tgreenwood: But I understand your concern and, to be honest I had the same one when I've just formatted the code and was looking for a way not to put it on separate lines, but in vain. By now I have accepted this way.

tgreenwood: Moreover, if we do not want lines longer than 120 characters, then what would be a preper way to format such objects? From my perspective, it is better to put each one on a separate line. I think it is easier to go throuh it looking got a property comparing to the way when properties are next to each other on two lines

@ewelinagr what do you think?

ewelinajozwik: > unfortunately, the only way to change it is to increase the max length, but 120 is already a big number

Hmm, ok, I though there would be a way to just ignore checks for such "class properties" and just allow these being wrapped to next line...

tgreenwood: @ewelinagr btw, in cases you want to keep custom formatting just put `// prettier-ignore` over the object/block

ewelinajozwik (approved):

PR #1503, opened 2024-03-14
Fix DOM text reinterpreted as HTML codeql error.

ewelinajozwik:


projects/mercury/src/file/fileUtils.js
```diff
@@ -142,7 +142,7 @@ export function generateUniqueFileName(fileName, usedNames = []) {
 
 export const decodeHTMLEntities = (htmlSource: string) => {
     const element = document.createElement('textarea');
-    element.innerHTML = htmlSource;
+    element.textContent = htmlSource;
```
tgreenwood, line 145: to be honest, now this looks weird: we create element -> set string to an attribute -> return the string value (actually the input param). Do we need this method or am I missing anything?

PR #1515, opened 2024-03-29
Devops/fairspc 23 deployment

tgreenwood:
Two jobs (workflows) added:

1. To package and push Helm Chart to repository (in GitHub). Triggered manually and by push to dev/release
2. To deploy Helm Chart of either input version or latest snapshot. Triggered manually only


.github/workflows/deploy_helm_chart.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Setup Helm
+        uses: azure/setup-helm@v4.1.0
+        with:
+          version: ${{ vars.HELM_VERSION }}
+
+      - name: Login with Helm
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ${{ vars.HELM_REGISTRY }} --username ${{ github.repository_owner }} --password-stdin
+
+      - name: Delete StatefulSet (if exists) to force re-deploy
```
ewelinajozwik, line 87: I think it would be good to check if the namespace exists first and if not - create it

.github/workflows/package_and_push_helm_chart.yaml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          VERSION=$(cat ./VERSION)
+          if [[ ${{ github.event.inputs.environment }} = "SNAPSHOT" ]] || [[ $BRANCH = "dev" ]]; then
+            VERSION=$VERSION-SNAPSHOT
+          fi
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+
+      - name: Push Helm chart to repository
+        run: |
+          helm package "charts/fairspace" --version ${{ env.VERSION }}
```
ewelinajozwik, line 51: before that it would be useful to run `helm dependency update <fairspace>` (as it is in the local deployment script: https://github.com/thehyve/fairspace/blob/dev/local-development/fairspace/deploy.sh#L48)

.github/workflows/deploy_helm_chart.yaml
```diff
@@ -0,0 +1,95 @@
...
+          location: europe-west1
+
+      - name: Setup Helm
+        uses: azure/setup-helm@v4.1.0
+        with:
+          version: ${{ vars.HELM_VERSION }}
+
+      - name: Login with Helm
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ${{ vars.HELM_REGISTRY }} --username ${{ github.repository_owner }} --password-stdin
```
ewelinajozwik, line 75: So there is no helm fetching or pulling specific chart needed anymore? 
(https://github.com/thehyve/fairspace?tab=readme-ov-file#fetch-charts)

.github/workflows/deploy_helm_chart.yaml
```diff
@@ -0,0 +1,95 @@
...
+          fi 
+
+      - name: Deploy Helm chart
+        run: |
+          helm upgrade --install fairspace oci://${{ vars.HELM_REGISTRY }}/fairspace \
+          --version ${{ env.VERSION }} \
+          --namespace=fairspace-${{ env.ENVIRONMENT }} \
+          --set-file saturn.vocabulary=./fns/fairspace/model/vocabulary.ttl \
+          --set-file saturn.views=./fns/fairspace/model/views.yaml \
+          --values ./fns/fairspace/deployment/${{ env.ENVIRONMENT }}/fairspace-test-values.yaml
```
ewelinajozwik, line 103: I've never used `--values`, we always pass the values path with `-f` flag. 
Does it for sure work? What is the difference between those two?

tgreenwood: image
They are equal, used it for the notation consistency (double-dash) and from hiegher-readability perspective, tested both, no difference found. But let me know if you don't like it, easily can switch to `-f`

.github/workflows/deploy_helm_chart.yaml
tgreenwood: Yep, no need to fetch it to deploy. Please see the doc reference:
image

But we need to login before upgrading or other operations for helm to be configured against a repository. For a reason, `helm repo add` is not applicable for GitHub packages as a repo

.github/workflows/deploy_helm_chart.yaml
ewelinajozwik: ok, good to know! I haven't seen it in [the helm values file](https://helm.sh/docs/chart_template_guide/values_files/) documentation,

.github/workflows/package_and_push_helm_chart.yaml
tgreenwood: we don't have dependencies, no need for now, but it is better to have in case one day I will have some - added the command, thanks for the point!

.github/workflows/deploy_helm_chart.yaml
tgreenwood: good point, added a step to check if ns exists. It will create a new one if it doesn't. Also, checked that delete stateful-set step works as expected (not failing if ns does not exists)

.github/workflows/deploy_helm_chart.yaml
```diff
@@ -0,0 +1,105 @@
...
+        run: |
+          # Check if the StatefulSet exists
+          if kubectl get statefulset fairspace --namespace fairspace-${{ env.ENVIRONMENT }} &> /dev/null; then
+            kubectl delete statefulset fairspace --namespace fairspace-${{ env.ENVIRONMENT }} --cascade=orphan
+            echo "StatefulSet fairspace deleted."
+          else
+            echo "StatefulSet fairspace does not exist."
+          fi 
+
+      - name: Check if namespace exists
```
ewelinajozwik: Shouldn't this step be there before deleting statefulset? 
Not sure if the statefulset step won't fail if trying to do `kubectl get statefulset fairspace --namespace ...` on non-existing namespace

.github/workflows/package_and_push_helm_chart.yaml
ewelinajozwik: on a really small note - not sure if the quotes are needed for `charts/fairspace`, but now there are no quotes for `dependency update` and they are there for `helm package`. Maybe good to make it consistent?

ewelinajozwik:
Since the helm version is being updated here and also there is a new place where the helm charts are stored, could you update our documentation? 
https://github.com/thehyve/fairspace/blob/dev/README.adoc#kubernetes-and-helm

.github/workflows/deploy_helm_chart.yaml
tgreenwood: this is exactly what I meant saying 'Also, checked that delete stateful-set step works as expected (not failing if ns does not exists)'. But no problem, swapped the steps.

ewelinajozwik: ah ok, sorry, I didn't notice that comment

tgreenwood:
@ewelinagr the doc has been updated, could you please review

.github/workflows/package_and_push_helm_chart.yaml
tgreenwood: no problem, added

README.adoc
```diff
@@ -2512,12 +2512,12 @@ Now everything should be ready to start using Fairspace:
 
 Requires:
 
-* Helm >= 3.13.x
-* kubectl >= 1.27.x
+* Helm >= 3.14.x
+* kubectl >= 1.26.x
```
ewelinajozwik: why version decrease? On GCP we are on 1.27.8

README.adoc
```diff
@@ -2572,13 +2572,11 @@ kubectl get ns
...
-===== Initialise helm and add fairspace repository
+===== Initialise helm and login to Fairspace GitHub Packages
 [source, shell]
 ----
-# Add the stable repo
-~/bin/helm/helm repo add stable https://charts.helm.sh/stable
-# Add the fairspace repo for reading
-~/bin/helm/helm repo add fairspace https://storage.googleapis.com/fairspace-helm
+# Add the stable repo (prepare and use personal access token, also known as PAT, within your GitHub account as a password)
+~/bin/helm/helm helm registry login ghcr.io/fairspace/helm-charts --username <your github account> --password-stdin
```
ewelinajozwik:
```suggestion
~/bin/helm/helm registry login ghcr.io/fairspace/helm-charts --username <your github account> --password-stdin
```

README.adoc
```diff
@@ -2693,7 +2691,7 @@ Create a new deployment (called _release_ in helm terminology) and
 install the Fairspace chart:
 [source, shell]
 ----
-~/bin/helm/helm install fairspace-new fairspace/fairspace --version 0.7.29 --namespace fairspace-new \
+~/bin/helm/helm install fairspace-new fairspace/fairspace --version 0.7.31 --namespace fairspace-new \
```
ewelinajozwik: we don't have version 0.7.31 released yet, what we have on dev is a snapshot, it will be released (and then updated in the documentation) when merging to `release` branch

README.adoc
tgreenwood: I referred to the compatibility table in  helm's doc :

image

ewelinajozwik: could be that helm still supports it, but I think it is better not to recommend a version that is end of life (https://endoflife.date/kubernetes)

README.adoc
tgreenwood: you are right, put 29 back

ewelinajozwik (approved):

PR #1519, opened 2024-04-17
Make the icons in application menu configurable (external services, metadata domains)

ewelinajozwik:
Implements [FAIRSPC-50](https://thehyve.atlassian.net/browse/FAIRSPC-50)

Unfortunately there is no way to include external files in helm install/upgrade command, so the current workaround is to pass icon files one by one with `--set-file` option (as I described in the readme file). Hopefully this feature will be included in helm, since there is already a PR for that:
  https://github.com/helm/helm/pull/10077 (unfortunately waiting for 4 years now...).

[FAIRSPC-50]: https://thehyve.atlassian.net/browse/FAIRSPC-50?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.adoc
```diff
@@ -1643,6 +1643,25 @@ curl -H 'Accept: application/json' 'http://localhost:8080/api/features/'
 ----
 ====
 
+===== Icons (SVG)
+
+|===
+| ``GET /api/iconsvg/{icon_name}``
```
tgreenwood, line 1649: do we want to put the restriction on icon format?

projects/mercury/src/common/IconAPI.js
```diff
@@ -0,0 +1,26 @@
...
+
+export type ObjectWithIconObjectUrl = ObjectWithIconPath & {
+    icon: string
+};
+
+export const getSvgIcon = (object: ObjectWithIconPath): Promise<ObjectWithIconObjectUrl> => {
+    if (object.iconPath === null) {
+        return Promise.resolve(object);
+    }
+    return axios.get(object.iconPath, {responseType: 'blob'}).then(response => {
```
tgreenwood: do I understand correctly, we don't cache now the icons? They are pretty static, something to make sense to cache in browser. What do you think?

projects/mercury/src/metadata/metadata-sources/InternalMetadataSourceContext.js
```diff
@@ -0,0 +1,35 @@
...
+import MetadataSourceContext from './MetadataSourceContext';
+import {DEFAULT_METADATA_VIEW_MENU_LABEL} from '../../constants';
+import {MetadataSource} from '../external-views/externalMetadataSourceUtils';
+
+const InternalMetadataSourceContext = React.createContext({});
+
+export const InternalMetadataSourceProvider = ({children}) => {
+    const {metadataSources = [], error, loading, refresh} = useContext(MetadataSourceContext);
+
+    const internalMetadataSource: MetadataSource = metadataSources.find(source => source.path === null);
```
tgreenwood, line 11: could you please explain why we accept the first one with path = null?

projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
```diff
@@ -44,11 +46,25 @@ public static class Storage {
         private String rootDirectoryIri;
     }
 
+    @Data
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public static class Service {
```
tgreenwood: I would name it more specifically, something like ExternalService, otherwise it is a bit general

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
```diff
@@ -0,0 +1,55 @@
...
+public class IconsResource {
+
+    private final PlutoConfig plutoConfig;
+    /**
+     * GET  /api/iconsvg/{icon_name} : returns an icon of specified name if exists.
+     */
+    @GetMapping(value = ICONS_PATH + "{icon_name}", produces = "image/svg+xml")
+    public ResponseEntity<byte[]> iconsvg(@PathVariable String icon_name) {
+        try {
+            InputStream in = getSvgIconInputStream(icon_name);
```
tgreenwood: stream closing is missed, to avoid leaking add the stream initialization to try () as InputStreams implement AutoClosable

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RestController;
+
+import nl.fairspace.pluto.config.dto.PlutoConfig;
+
+import static nl.fairspace.pluto.config.Urls.ICONS_PATH;
+
+@RestController
+@Slf4j
+@AllArgsConstructor
```
tgreenwood: Resource is a singleton and unmodifiable. This is correct to define fields final, `@RequiredArgsConstructor` can be used instead

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
```diff
@@ -0,0 +1,55 @@
...
+public class IconsResource {
+
+    private final PlutoConfig plutoConfig;
+    /**
+     * GET  /api/iconsvg/{icon_name} : returns an icon of specified name if exists.
```
tgreenwood: `@return` tag can be used for proper layout and common practice

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
```diff
@@ -0,0 +1,55 @@
...
+                return ResponseEntity.notFound().build();
+            }
+            return ResponseEntity.ok(IOUtils.toByteArray(in));
+        } catch (IOException e) {
+            log.error("Error reading image", e);
+            return ResponseEntity.notFound().build();
+        }
+    }
+
+    public String getIconUrl(String name) {
```
tgreenwood: If to leave it here, I would suggest to make it private as it is for internal usage. But rather I would suggest to make PlutoConfig responsible for the methods and exposing them. Resource/Controller layer should be responsible only for accepting a request and routing it further to dedicated service

projects/pluto/src/main/java/nl/fairspace/pluto/web/MetadataSourcesResource.java
```diff
@@ -33,7 +34,8 @@ public ResponseEntity<List<MetadataSourceInfo>> metadataSources() {
                 .map(source -> new MetadataSourceInfo(
                         source.getName(),
                         source.getLabel(),
-                        String.format("/api/metadata-sources/%s", source.getName())))
+                        source.getName() != null ? String.format("/api/metadata-sources/%s", source.getName()) : null,
```
tgreenwood: I would put the parametrized path to const

README.adoc
ewelinajozwik: I did that due to the fact that svg need to be parsed differently than e.g. png/jpg. Svg is just text (image/xml+svg), any other format needs to be decoded/encoded.
For any other format I would add a separate endpoint (if needed) - there would be a different code on the front-end side needed for these anyway.

projects/mercury/src/common/IconAPI.js
ewelinajozwik: Yes, you're right, I had a plan to add caching, but then forgot about it! Thanks, I'll add it!

projects/mercury/src/metadata/metadata-sources/InternalMetadataSourceContext.js
ewelinajozwik: I described it in the documentation - path is the url of"|Fairspace instance to connect to. If the url is not specified, the metadata source will be treated as the internal one.
| *Important!* There should only be a single configuration of internal metadata (only the first one will not be ignored)."

projects/pluto/src/main/java/nl/fairspace/pluto/config/dto/PlutoConfig.java
ewelinajozwik: I can try different name... This is how we call this option in the configuration, just `services`, and I just moved it from Saturn to Pluto.

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
ewelinajozwik: true, I missed the closing part, thanks!

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
ewelinajozwik: Does it matter in this case if I use RequiredArgsConstructor vs AllArgsConstructor?

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
ewelinajozwik: This one is not only used internally and PlutoConfig is more of a DTO, so I don't think methods belong there. I will see how to organize this better

frankyhollywood:
Overall looks good!

The only thing I don't get is why services configuration is moved from Saturn to Pluto, can you explain a bit because it makes it more complex than it used to be.

ewelinajozwik:
> Overall looks good!
> 
> The only thing I don't get is why services configuration is moved from Saturn to Pluto, can you explain a bit because it makes it more complex than it used to be.

My goal was to simplify things actually :) 
"Services" were added before Pluto was introduced, that is why these were in Saturn. 
What this endpoint does is just passing the defined configuration to Mercury, no logic there - it is the same as `/storages`, `/metadata-sources` and `/config` endpoints that are already in Pluto. 
And I wanted to place icons together with Pluto, since it is already responsible for serving the static content (Services and icons are connected).

projects/mercury/src/common/IconAPI.js
ewelinajozwik: It is added now

projects/pluto/src/main/java/nl/fairspace/pluto/web/IconsResource.java
tgreenwood: not critical,  the only advantage is that it will be safer to add new dependencies

frankyhollywood (approved):

tgreenwood (approved):

PR #1520, opened 2024-04-24
Remove code for special decoding of webdav response.

ewelinajozwik:
The `decodeHTMLEntities` seems not to be needed and it introduces vulnerability: "DOM text reinterpreted as HTML" -  CWE-79, CWE-116.

There is an issue with special characters in external storages, but this function doesn't fix the problem (reported in [FAIRSPC-62](https://thehyve.atlassian.net/browse/FAIRSPC-62)) 

[FAIRSPC-62]: https://thehyve.atlassian.net/browse/FAIRSPC-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/mercury/src/file/FileAPI.js
+++ b/projects/mercury/src/file/FileAPI.js
@@ -2,14 +2,7 @@ import {createClient} from 'webdav';
 import qs from 'qs';
 import {compareBy, comparing} from '../common/utils/genericUtils';
 // eslint-disable-next-line import/no-cycle
-import {
-    decodeHTMLEntities,
-    encodePath,
-    generateUniqueFileName,
-    getFileName,
-    joinPaths,
-    joinPathsAvoidEmpty
-} from './fileUtils';
+import {encodePath, generateUniqueFileName, getFileName, joinPaths, joinPathsAvoidEmpty} from './fileUtils';
 import {handleHttpError} from '../common/utils/httpUtils';
 
 // Ensure that the client passes along the credentials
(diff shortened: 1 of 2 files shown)


projects/mercury/src/file/FileAPI.js
```diff
@@ -479,12 +472,6 @@ class FileAPI {
     mapToFile = fileObject => {
         const properties = {...fileObject, ...(fileObject.props || {})};
         delete properties.props;
-        Object.keys(properties).forEach(key => {
-            // The WebDAV client does not properly decode the XML response,
```
tgreenwood, line 483: Let me be honest, I still don't understand this part of the code. Did you check the problem this piece of code solves? Don't we remove a fix of something?

tgreenwood: I mean I see your comment, I saw the ticket you created. Does this ticket cover the issue? How critical is it?

ewelinajozwik: That is the thing, it doesn't seem to be solving any problem. Maybe there was something in previous versions of webdav, but now with or without it makes no difference (I tried various properties, various special characters - still no difference if it is there or not)

tgreenwood: good, let's indeed remove it and see how it goes

tgreenwood (approved):

PR #1523, opened 2024-05-23
Fix numeric facet slider tooltip.

ewelinajozwik:
Fixes [FAIRSPC-9](https://thehyve.atlassian.net/browse/FAIRSPC-9)

[FAIRSPC-9]: https://thehyve.atlassian.net/browse/FAIRSPC-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/mercury/src/metadata/views/facets/NumericalRangeSelectionFacet.js
+++ b/projects/mercury/src/metadata/views/facets/NumericalRangeSelectionFacet.js
@@ -126,7 +126,7 @@ const NumericalRangeSelectionFacet = (props: MetadataViewFacetProperties) => {
     };
 
     const renderInput = (inputValue, onInputChange, placeholder) => (
-        <ValueLabelComponent value={nonEmptyNumber(inputValue, '')}>
+        <ValueLabelComponent value={nonEmptyNumber(inputValue, placeholder)}>
             <Input
                 value={nonEmptyNumber(inputValue, '')}
                 margin="dense"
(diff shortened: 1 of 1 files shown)


tgreenwood:
That's impressive how quickly you fixed that! The only moment I found is that the tooltip is not shown until you change the pre-set long value

image

ewelinajozwik:
> The only moment I found is that the tooltip is not shown until you change the pre-set long value

@tgreenwood It is fixed now - if no value is specified, the tooltip will show the placeholder value

tgreenwood:
All is good, I am just thinking about one more tiny point
image


now the tooltip is available like every time user hovers cursor over the edit field, even when there is no need for that (small numbers). Is it something we could hide while no need to duplicate the info for users to make it visible? I would also understand, if you called my comment "an overhead' :)

ewelinajozwik:
> now the tooltip is available like every time user hovers cursor over the edit field, even when there is no need for that (small numbers). Is it something we could hide while no need to duplicate the info for users to make it visible? I would also understand, if you called my comment "an overhead' :)

hmm, not sure how to do that :sweat_smile: Depending on things like screen resolution, zoom etc the number of digits would be different... Not sure if having the tooltip always there would be now considered annoying

tgreenwood:
never

> > now the tooltip is available like every time user hovers cursor over the edit field, even when there is no need for that (small numbers). Is it something we could hide while no need to duplicate the info for users to make it visible? I would also understand, if you called my comment "an overhead' :)
> 
> hmm, not sure how to do that 😅 Depending on things like screen resolution, zoom etc the number of digits would be different... Not sure if having the tooltip always there would be now considered annoying

nevermind then, let's have it as is

tgreenwood (approved):

PR #1524, opened 2024-05-27
FAIRSPC-68: fixed date filter for Jena

tgreenwood:
The bugfix found on a datamodel where we define some attributes as date fields. The whole application operates with datetime format (front-end with its Date type, back-end with Instant/timestamp for Postrgres) while in Jena we use date format. But all  mappings for Jena map data type to date time. It wouldn't be an easy way to switch to pure date format everywhere (especially because we do not support data model evolution within data model versioning). 

I suggest then just to add handling datetime for Jena in one place instead of reworking it in multiple places with un-answered question about how easily to deploy it to already existing environments with datetime formats.

### DIFF
--- a/projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
@@ -364,7 +364,9 @@ public class SparqlQueryService implements QueryService {
             case Identifier, Term, TermSet -> makeNode(createURI(o.toString()));
             case Text, Set -> makeString(o.toString());
             case Number -> makeDecimal(o.toString());
-            case Date -> makeDateTime(convertDateValue(o.toString()));
+                // Extract date only as it comes in format "yyyy-MM-ddTHH:mm:ss.SSSX"
+                // Leave it as is in case of adding new DateTime filter format
+            case Date -> makeDate(o.toString().split("T")[0]);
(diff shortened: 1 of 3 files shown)


projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
```diff
@@ -364,7 +364,9 @@ private static NodeValue toNodeValue(Object o, ColumnType type) {
             case Identifier, Term, TermSet -> makeNode(createURI(o.toString()));
             case Text, Set -> makeString(o.toString());
             case Number -> makeDecimal(o.toString());
-            case Date -> makeDateTime(convertDateValue(o.toString()));
+                // Extract date only as it comes in format "yyyy-MM-dd'T'HH:mm:ss.SSSX"
+                // Leave it as is in case of adding new DateTime filter format
+            case Date -> makeDate(o.toString().split("T")[0]);
```
ewelinajozwik, line 369: I haven't tested it, but wouldn't it leave `yyyy-MM-dd'` after the split (with an apostrophe `'` at the end)? 

But it looks like the format that it actually comes in is without `'` `'` around `T`, right?

tgreenwood: It's confusing, agree. I removed the quotes in the comment as request does not contain them. It was an unlucky attempt to highlight

ewelinajozwik (approved):

frankyhollywood (approved):

PR #1529, opened 2024-06-17
Add initial version of a new Fairspace 2.0 user interface

ewelinajozwik:


tgreenwood:
An impressive work!!! No way I can find something in the css files, looks good from a glance. Let me just check the UI look.

Found buttons stick to each other
image

tgreenwood:
image

please out it into upper case - 'back'

tgreenwood:
What do you think if we add a non-zero margin to scroll?
image

Now the scrollbar is a bit cropped in the edge positions

tgreenwood:
One more point about scroll bars: from my perspective it is difficult to click on the scroll bar, it requires to be very accurate in your clicks. Can we make the hover area a it bigger?

tgreenwood:
Point to discuss about scroll bar color: this is personal, but let me share - scroll bars color is a bit out of the color palette, it would be more in harmony with the whole style if it was of the facet drop down 
image

OR like the main one

image



What ado you think?

tgreenwood:
image
It looks like a css bug when we select the first tab. Any ideas on how we could highlighted in a different way?

tgreenwood:
Why do we need the horizontal scroll bar on the metadata details popup? It breaks the beauty of the rounded form :(
image

tgreenwood:
image

and 
image

looks like upper margin got lost

tgreenwood:
the tooltip is got hidden again

image

ewelinajozwik:
> the tooltip is got hidden again
> 
> image

The branch is not fully up-to-date with dev, the fix is not included.

tgreenwood:
image
The dark rectangle is not rounded as all the interface components

tgreenwood:
image
Looks like broken layout

tgreenwood:
image
icaons are not well visible, maybe make them of same color as the text?

tgreenwood (approved):

PR #1530, opened 2024-06-17
Llm search rebased on new GUI

frankyhollywood:
Added LLM search to Fairspace


projects/mercury/public/public/images/logo_white.png
frankyhollywood, line 1: it's in folder 'public/public/...', is duplicate public on purpose?

PR #1531, opened 2024-06-17
Llm search 

frankyhollywood:


projects/mercury/src/llm/Conversation.styles.js
```diff
@@ -0,0 +1,154 @@
+import * as consts from '../constants';
+
+const styles = theme => ({
```
tgreenwood, line 3: shouldn't we keep it in css file instead?

projects/mercury/src/llm/FulltextAPI.js
```diff
@@ -0,0 +1,68 @@
+import axios from 'axios';
+import {extractJsonData, handleHttpError} from '../common/utils/httpUtils';
+
+export const HEADERS = {'Content-Type': 'application/json', Accept: 'application/json'};
+
+/**
+ * Search for resources based on name or description, given query as a simple text.
+ */
+class FulltextAPI {
+    remoteURL = '/api/aisearch/';
```
tgreenwood, line 10: I would put 'const' here

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+                }).collect(Collectors.toList());
+
+                var result = new ArrayList<JSONObject>();
+                for (var c : content) {
+                    try {
+                        var obj = new JSONObject(c);
+                        var conversation = new JSONObject();
+                        conversation.put("id", obj.getJSONObject("conversation").getString("conversationId"));
+                        conversation.put("topic", obj.getJSONObject("conversation").getJSONArray("messages").getJSONObject(0).getJSONObject("userInput").getString("input"));
+                        conversation.put("start", new LlmConversation().getStartTime(obj.getJSONObject("conversation").getString("startTime")));
```
tgreenwood: Please, move all logic to another layer, 'service' not to mixup controller layer with logic and for easier logic testing. 

Also, I would create instance of LlmConversation on the class instance creation instead of creating it on each request

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+public class AiSearchApp extends BaseApp {
+    final String CACHE_DIR = "./data/conversations/";
+
+    public AiSearchApp(String basePath) {
+        super(basePath);
+    }
+
+    @Override
+    protected void initApp() {
+        get("/newchat", "application/json", (req, res) -> {
```
tgreenwood, line 32: I would think of the API in  REST/HTTP API way:

- use HTTP DELETE method for deletion
- use HTTP POST method for any type of creation (with the GET '/newchat' we change the state, it is not even idempotent)
- not to use verbs in the path, use proper HTTP methods 


We could consider API like this:

- GET /chat - get all conversations
- POST /chat - current /newchat
- GET /chat/{id} - current /history/:id NOTE: I did not gat the difference between `'/history/:id'` and `'/conversation/:id'`, could you elaborate please?
- POST /chat/{id}/query
- DELETE /chat/{id}

There is one more in the code POST '/' - I don't get it, please elaborate too

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+
+                return searchResult;
+            } catch (Exception e) {
+                return handleError(req, e);
+            }
+        });
+    }
+
+    private Object handleError(Request req, Exception e) {
+        log.error("Unexpected error in AiSearchApp: " + e.getMessage() + "\n\nrequest: " + req.body());
```
tgreenwood, line 220: I wouldn't put class name in the log as the name may be changed, logger configuration is the right place for that

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private String getUserKey() {
+        var token = getAccessToken();
+
+        if (token.getIssuedFor() != null && token.getPreferredUsername() != null) {
+            return Math.abs(token.getIssuedFor().hashCode()) + "@"
+                    + Math.abs(token.getPreferredUsername().hashCode());
+        }
+
+        throw new UnsupportedOperationException("User key not found in token");
```
tgreenwood, line 232: UnsupportedOperationException is used to literally say that something is not supported, should be another, at least RuntimeException and a proper wrapper

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+        });
+    }
+
+    private Object handleError(Request req, Exception e) {
+        log.error("Unexpected error in AiSearchApp: " + e.getMessage() + "\n\nrequest: " + req.body());
+        return "{}";
+    }
+
+    private String getUserKey() {
```
tgreenwood, line 224: I would suggest not to bring this method here not to violate single responsibility principle, rather it could be another method of RequestContext class

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+            if (!Files.exists(Paths.get(CACHE_DIR + getUserKey()))) {
+                return new ArrayList<JSONObject>();
+            }
+
+            try (var conversations = Files.list(Paths.get(CACHE_DIR + getUserKey()))) {
+                var content = conversations.map(path -> {
+                    try {
+                        return new String(Files.readAllBytes(path));
+                    } catch (Exception e) {
+                        return "{}";
```
tgreenwood: looks like error handling missed here

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+                    }
+                }).collect(Collectors.toList());
+
+                var result = new ArrayList<JSONObject>();
+                for (var c : content) {
+                    try {
+                        var obj = new JSONObject(c);
+                        var conversation = new JSONObject();
+                        conversation.put("id", obj.getJSONObject("conversation").getString("conversationId"));
```
tgreenwood, line 62: I would suggest to use POJO classes as more Java way instead of manipulation over pure JSON, it would also mitigate appearance of potential errors

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
```diff
@@ -0,0 +1,222 @@
...
+                var result = new LlmConversation().continueChat(conversationId, query);
+                var filepath = CACHE_DIR + getUserKey();
+                var file = new java.io.File(filepath + "/" + conversationId + ".json");
+
+                Files.createDirectories(Paths.get(filepath));
+
+                try (FileWriter writer = new FileWriter(file)) {
+                    writer.write(result.toString());
+                } catch (Exception e) {
+                    System.out.println("Error writing conversation history file: " + e.getMessage());
```
tgreenwood, line 161: please use logger, not system output

tgreenwood: also, check other places, there are 6 of them

ewelinajozwik:
To me the whole AISearch API belongs in Pluto, not Saturn. It is mainly based on redirection, this is one of the reasons Pluto was introduced for. Also, we already have ExternalSearch logic, which is really similar to this and stays within Pluto. To me Saturn should be all about our internal storage, not redirecting to external services

projects/saturn/src/main/java/io/fairspace/saturn/services/search/AiSearchApp.java
tgreenwood, line 220: hm, we don't let user know that something went wrong, let's re-though it and handle with proper http error code in the handler in BaseApp

projects/saturn/build.gradle
```diff
@@ -69,6 +69,11 @@ dependencies {
     implementation "org.postgresql:postgresql:${postgresqlVersion}"
     implementation "com.zaxxer:HikariCP:5.1.0"
 
+    implementation("org.json:json:20240303")
+
+    // LLM is customer specific. By default the jar will be a demo implementation, overwrite it with customer specific implementation
+    implementation files("libs/nl-hyve-fairspace-demo-v2.4.jar")
```
tgreenwood, line 75: To be honest, I am still not happy with the approach :( We are adding not only the dependency version, but artifact itself too. So, it is pretty tough to change it, too tight coupling. Also, let me know if I am stupid, I am not getting how we can conditionally AskAI switch on and off.

tgreenwood:
also I remember Ask AI is a POC (it also does not include any tests), but way are we adding it to the dev branch?

tgreenwood:
> To me the whole AISearch API belongs in Pluto, not Saturn. It is mainly based on redirection, this is one of the reasons Pluto was introduced for. Also, we already have ExternalSearch logic, which is really similar to this and stays within Pluto. To me Saturn should be all about our internal storage, not redirecting to external services

There is a bit of logic around managing the chats and chat history, not sure it would be correct to put business logic to the gateway layer. I still believe it should be a separate service, not a part of Pluto nor Saturn. It would also let us easily toggle the whole feature

projects/mercury/src/llm/Conversation.styles.js
frankyhollywood: In FS we use styles.js files.

ewelinajozwik:
Could you also make the naming more consistent? Especially the clases are currently named from LLM search, Fulltext search, AI search to just "conversation". I also see references like "Article search". I would prefix everything "AI" or "Llm" and then decide on single name afterwards, preferably "AISearch...", since this is how the API endpoint is named now. Then instead of:

= llm
== Conversation.js
== FulltextAPI.js
== LlmSearchPage.js

you would have:

= ai
== AISearchPage.js
== AISearchAPI.js
== AISearchConversation.js or AISearchPanel.js

projects/mercury/src/llm/Conversation.js
```diff
@@ -0,0 +1,415 @@
...
+        if (query === '') {
+            return;
+        }
+
+        performSearch()
+            .then(processResponseMessages)
+            .then(() => setLoading(false))
+            .catch(() => {
+                handleHttpError('Connection error.');
+                handleSearchError('Fairspace article search is not available at the moment');
```
ewelinajozwik, line 202: "Fairspace article search" - is this always articles? If so, we should make it more specific, otherwise we need more generic naming

projects/mercury/src/llm/Conversation.js
```diff
@@ -0,0 +1,415 @@
...
+        if (webResult && webResult.length > 0) {
+            setArticleIri(webResult[0].id);
+        } else {
+            setArticleIri('');
+        }
+    };
+
+    const showArticle = articleId => {
+        if (articleId !== null && articleId.length > 0) {
+            LocalSearchAPI.lookupSearch(articleId, 'https://www.fns-cloud.eu/study').then(extractArticleIri);
```
ewelinajozwik, line 216: I don't like hardcoded prefixes in the generic UI, this is a component that works in a really specific setup...

projects/mercury/src/llm/Conversation.js
```diff
@@ -0,0 +1,415 @@
...
+                    })}
+            </div>
+        );
+    };
+
+    const renderArticleReferences = () => {
+        return (
+            <div>
+                {responseArticles && responseArticles.length > 0 && (
+                    <Typography variant="caption">Source studies:</Typography>
```
ewelinajozwik, line 277: it assumes a model with "study" entities?

projects/mercury/src/llm/Conversation.js
```diff
@@ -0,0 +1,415 @@
...
+import CloseIcon from '@mui/icons-material/Close';
+import SearchIcon from '@mui/icons-material/Search';
+import withStyles from '@mui/styles/withStyles';
+import FulltextAPI from './FulltextAPI';
+import styles from './Conversation.styles';
+import {handleHttpError} from '../common/utils/httpUtils';
+import LinkedDataEntityPage from '../metadata/common/LinkedDataEntityPage';
+import {LocalSearchAPI} from '../search/SearchAPI';
+
+const Conversation = props => {
```
ewelinajozwik, line 28: This huge class should be divided into smaller modules, currently it is hard to follow what happens here

PR #1534, opened 2024-06-19
FAIRSPC-102: fixed view update on triple deletion (property deletion)

tgreenwood:
For a reason all not present property triples in RDF entity were filtered out before entity update in RDF


projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewStoreClient.java
```diff
@@ -203,17 +203,13 @@ public void addLabel(String id, String type, String label) throws SQLException {
     public int updateRows(String view, List<Map<String, Object>> rows, boolean bulkInsert) throws SQLException {
         var viewTable = configuration.viewTables.get(view);
         var config = configuration.viewConfig.get(view);
-        // Find the columns for which there are values in the rows
+        // Find the columns in the rows of type different from Set
         var columnNames = rows.stream()
-                .flatMap(row -> row.entrySet().stream()
-                        .filter(entry -> entry.getValue() != null)
-                        .map(Map.Entry::getKey))
+                .flatMap(row -> row.keySet().stream())
```
frankyhollywood, line 208: you removed a null validation, is that correct?

frankyhollywood (approved):

PR #1547, opened 2024-08-08
Fairspc 108 metadata page count

tgreenwood:


projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
```diff
@@ -1,18 +1,30 @@
...
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.*;
 import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.dataformat.yaml.*;
 import jakarta.validation.constraints.*;
 
 public class ViewsConfig {
+
+    private Map<String, View> viewConfig;
```
ewelinajozwik: Could you rename this to something like `viewConfigMap` or `nameToViewConfigMap`? 
It is confusing that `getViewConfig` gives you `View`, but `viewConfig` is a map of names to `Views`.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
```diff
@@ -169,14 +195,10 @@ private ValueDTO toValueDTO(RDFNode node) {
...
-    private Query getQuery(CountRequest request) {
+    private Query getQuery(CountRequest request, boolean isCount) {
         var view = getView(request.getView());
 
-        var builder = new StringBuilder("PREFIX fs: <")
-                .append(FS.NS)
-                .append(">\n\nSELECT ?")
-                .append(view.name)
-                .append("\nWHERE {\n");
+        var builder = new StringBuilder().append("SELECT ").append("%s").append("\nWHERE {\n");
```
ewelinajozwik: why this change? Why not putting the view name already here, instead of the '%s'? And why adding the prefix later?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
```diff
@@ -376,23 +419,25 @@ private static boolean convertBooleanValue(String value) {
...
+                    return new CountDTO(-1, false);
                 }
-                return new CountDTO(count, false);
-            } catch (QueryCancelledException e) {
-                return new CountDTO(count, true);
-            }
-        });
+            });
+        } catch (QueryCancelledException e) {
+            return new CountDTO(-1, true);
```
ewelinajozwik: SparqlQueryService and JdbcQueryService return different values in this case now. Previously both returned 0, now Sparql returns -1, Jdbc 0.

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewConfigDto.java
```diff
@@ -0,0 +1,15 @@
+package io.fairspace.saturn.services.views;
+
+import lombok.Value;
+
+import io.fairspace.saturn.config.ViewsConfig;
+
+@Value
+public class ViewConfigDto {
```
ewelinajozwik: This class is not used anywhere...

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
ewelinajozwik, line 1: Please replace line 174 with the new function `viewsConfig.getViewConfig(j.view)`

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -189,7 +189,11 @@ protected List<ViewDTO> fetchViews() {
                             columns.add(new ColumnDTO(joinView.name + "_" + c.name, c.title, c.type, j.displayIndex));
                         }
                     }
-                    return new ViewDTO(v.name, v.title, columns);
+                    Long maxDisplayCount = viewsConfig
+                            .getViewConfig(v.name)
+                            .map(t -> t.maxDisplayCount)
+                            .orElse(null);
+                    return new ViewDTO(v.name, v.title, columns, maxDisplayCount);
```
ewelinajozwik: why retrieving maxDisplayCount again, if you have it available as `v.maxDisplayCount`?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
ewelinajozwik: ok, I understand now, but the query building logic is now hard to follow...

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
tgreenwood: Agree, now it is different. In turn, I am not sure that 0 is the right value in case of something went wrong as potentially it is valid count value while -1 is not valid highlighting that something went wrong. Will put 0 for now

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: we extract the value if none null ViewConfig found by name

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
tgreenwood: agree

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewConfigDto.java
tgreenwood: good catch

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
ewelinajozwik: but you are already iterating through views and you can have that value here from `v.maxDisplayCount`, can't you?

ewelinajozwik:
```suggestion
                    return new ViewDTO(v.name, v.title, columns, v.maxDisplayCount);
```

tgreenwood: haha, indeed, facepalm

tgreenwood: applied

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewConfigDto.java
tgreenwood: removed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
tgreenwood: put 0

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
tgreenwood: renamed

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
tgreenwood: agree, slightly improved, but the prefix to be added in the end anywhere due to potential query wrapping for count query. What do you think now?

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
```diff
@@ -189,7 +186,7 @@ protected List<ViewDTO> fetchViews() {
                             columns.add(new ColumnDTO(joinView.name + "_" + c.name, c.title, c.type, j.displayIndex));
                         }
                     }
-                    return new ViewDTO(v.name, v.title, columns);
```
ewelinajozwik, line 192: Could you update the documentation (updates to the API endpoint)?

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
```diff
@@ -68,6 +80,12 @@ public static class View {
          * The name of the items that appear as rows.
          */
         public String itemName;
+        /**
+         * The max count value to be requested for a view total count.
+         * If total count is greater than this max value, the total count value will look like 'more than 1000' on FE.
+         * This is to prevent performance issues when the total count is too large.
+         */
+        public Long maxDisplayCount;
```
ewelinajozwik, line 88: Maybe with @Positive annotation? Then it needs to be > 0 or null.
```suggestion
        @Positive
        public Long maxDisplayCount;
```
We don't validate enough the values of this config, there is a ticket to improve it, but we can already start making it better. I guess the negative value of limit in Postgres doesn't make sense

projects/saturn/src/main/java/io/fairspace/saturn/services/views/SparqlQueryService.java
ewelinajozwik: sure, looks good!

projects/saturn/src/main/java/io/fairspace/saturn/config/ViewsConfig.java
tgreenwood: I would do it in the scope of this ticket https://thehyve.atlassian.net/browse/FAIRSPC-80. Now there are a few annotations like @NotBlank in the properties class, but they don't work, validation is not triggered anywhere. So, I would leave it for now, but I dropped it in the 80th ticket description

projects/saturn/src/main/java/io/fairspace/saturn/services/views/ViewService.java
tgreenwood: good point, done

README.adoc
```diff
@@ -1239,7 +1239,7 @@ curl -X POST -H 'Content-type: application/json' -H 'Accept: application/json' -
 |===
 3+| ``POST /api/views/count``
 
-3+| Count rows of a view matching request filters.
+3+| Count rows of a view matching request filters. If `maxDisplayCount` configured in the `views.ttl` for a view, then the count for the view is limited by this value if total count exceeds it. Otherwise, the total count is returned.
```
ewelinajozwik: `views.yaml`, not `views.ttl`

tgreenwood: oops, fixed

ewelinajozwik (approved):

PR #1553, opened 2024-08-27
Fix vulnerabilities in node packages.

ewelinajozwik:
According to `yarn audit` this solves 33 vulnerabilities. There are 3 left that require more changes:
- 2 of them (moderate and high level) are introduced by `react-scripts` - react scripts are no longer activly maintained, so we should find an alternative
- 1 of them (moderate) introduced by `webdav` package - unfortunately in webdav-client version >= 5 there was a switch to fetch from axios - this doesn't work well for us in case of file uploads (issues caused on the backed side) - see https://github.com/perry-mitchell/webdav-client/issues/352, also fetch doesn't allow tracking uploads. We need to investigate switching to version >=5 (impact of fetch for the webdav related oprations) or consider switching to other package instead of https://www.npmjs.com/package/webdav

### DIFF
--- a/projects/mercury/package.json
+++ b/projects/mercury/package.json
@@ -31,7 +31,7 @@
     "@mui/x-date-pickers": "^5.0.11",
     "@testing-library/react-hooks": "^8.0.1",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.8.0",
-    "axios": "^1.6.7",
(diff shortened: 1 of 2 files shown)


tgreenwood (approved):

tgreenwood:
Thanks for details, let's plan the investigation. And good job with the fix!

PR #1554, opened 2024-08-30
Fix file /api/search endpoint when collections not configured as a postgres view.

ewelinajozwik:
Fixes FAIRSPC-106

### DIFF
--- a/projects/saturn/src/main/java/io/fairspace/saturn/services/search/JdbcFileSearchService.java
+++ b/projects/saturn/src/main/java/io/fairspace/saturn/services/search/JdbcFileSearchService.java
@@ -1,0 +1,49 @@
+package io.fairspace.saturn.services.search;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import io.milton.resource.CollectionResource;
+import lombok.SneakyThrows;
+import lombok.extern.log4j.Log4j2;
+
+import io.fairspace.saturn.config.Config;
+import io.fairspace.saturn.config.ViewsConfig;
+import io.fairspace.saturn.rdf.transactions.Transactions;
+import io.fairspace.saturn.services.views.ViewStoreClientFactory;
+import io.fairspace.saturn.services.views.ViewStoreReader;
+
+import static io.fairspace.saturn.webdav.PathUtils.getCollectionNameByUri;
+
+@Log4j2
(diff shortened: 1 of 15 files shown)


projects/saturn/src/main/java/io/fairspace/saturn/services/search/JdbcFileSearchService.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Log4j2
+public class JdbcFileSearchService implements FileSearchService {
+    private final Transactions transactions;
+    private final CollectionResource rootSubject;
+    private final Config.Search searchConfig;
+    private final ViewsConfig viewsConfig;
+    private final ViewStoreClientFactory viewStoreClientFactory;
+
+    public JdbcFileSearchService(
```
tgreenwood, line 26: you can use @RequiredArgsContructor by lombok instead

projects/saturn/src/main/java/io/fairspace/saturn/services/search/JdbcFileSearchService.java
```diff
@@ -0,0 +1,49 @@
...
+            Transactions transactions,
+            CollectionResource rootSubject) {
+        this.searchConfig = searchConfig;
+        this.viewStoreClientFactory = viewStoreClientFactory;
+        this.transactions = transactions;
+        this.rootSubject = rootSubject;
+        this.viewsConfig = viewsConfig;
+    }
+
+    @SneakyThrows
```
tgreenwood, line 39: why do you use SneakyThrows? Also, try block does not catch anything, I would rather catch there, log it and re-throw unchecked exception if needed

ewelinajozwik: For this change I focused only on fixing the bug, which required restructuring the code. The code however is the same as it was (I only copy-pasted it to a different class). I didn't focus on the improvements - SneakyThrows was just there. If it needs to be changed, more classes needs to be refactored, same for the RequiredArgsContructor - for consistency we should change it everywhere then.

tgreenwood (approved):
No need to improve in the scope of this ticket for sure, missed that it is just copy/paste. In this scope all look good to me

PR #1567, opened 2024-10-11
Refactoring/fairspc 81

tgreenwood:


projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
```diff
@@ -0,0 +1,58 @@
...
+import io.fairspace.saturn.config.Services;
+import io.fairspace.saturn.controller.dto.CountDto;
+import io.fairspace.saturn.controller.dto.FacetsDto;
+import io.fairspace.saturn.controller.dto.ViewPageDto;
+import io.fairspace.saturn.controller.dto.ViewsDto;
+import io.fairspace.saturn.controller.dto.request.CountRequest;
+import io.fairspace.saturn.controller.dto.request.ViewRequest;
+
+@RestController
+@RequestMapping("${application.basePath}/views")
```
ewelinajozwik: For some controllers you have "/api/", for some "${application.basePath}", is there a reason for that?

projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @GetMapping(value = "/", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ViewsDto getViews() {
+        var views = services.getViewService().getViews();
+        return new ViewsDto(views);
+    }
+
+    @PostMapping(value = "/", consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<ViewPageDto> createView(@Valid @RequestBody ViewRequest requestBody) {
```
ewelinajozwik: createView is confusing, maybe "queryView" or "getViewBy..."

projects/saturn/src/main/java/io/fairspace/saturn/services/workspaces/WorkspaceService.java
```diff
@@ -38,41 +38,51 @@ public WorkspaceService(Transactions tx, UserService userService) {
...
+                                .filter(ws -> userService.currentUser().isCanViewPublicMetadata()
+                                        || ws.isCanManage()
+                                        || ws.isCanCollaborate())
+                                .collect(toList());
+            } catch (Throwable e) {
+                log.error("+++++++++++++++++++++++++++++");
+                log.error("+++++++++++++++++++++++++++++");
+                log.error("+++++++++++++++++++++++++++++");
+                log.error("+++++++++++++++++++++++++++++");
+                log.error("+++++++++++++++++++++++++++++");
```
ewelinajozwik: Why such a strange log message?

projects/saturn/src/main/java/io/fairspace/saturn/services/workspaces/WorkspaceService.java
```diff
@@ -38,41 +38,51 @@ public WorkspaceService(Transactions tx, UserService userService) {
...
+                log.error("Error listing workspaces", e);
+                throw e;
```
ewelinajozwik: is that handled correctly within the transaction?

projects/saturn/src/main/resources/application.yaml
```diff
@@ -35,6 +35,7 @@ jwt:
       principal-attribute: preferred_username
 
 application:
+  basePath: /api
```
ewelinajozwik: "/api" hardcoded value not replaced in all the places with this variable

projects/saturn/src/main/java/io/fairspace/saturn/controller/UserController.java
```diff
@@ -0,0 +1,36 @@
...
+import io.fairspace.saturn.services.users.UserService;
+
+@RestController
+@RequestMapping("/api/users/")
+@RequiredArgsConstructor
+public class UserController {
+
+    private final UserService userService;
+
+    @GetMapping
```
ewelinajozwik: The annotations are not consistent in all controllers, especially for getters - sometimes you specify only `@GetMapping`, sometimes it's more extensive including: `@GetMapping(value = "/", produces = MediaType.APPLICATION_JSON_VALUE)`

ewelinajozwik:
Please also update log4j.properties - it still has some spark specific settings

projects/saturn/src/main/java/io/fairspace/saturn/controller/dto/SearchResultDto.java
```diff
@@ -1,12 +1,12 @@
...
+package io.fairspace.saturn.controller.dto;
 
 import lombok.Builder;
 import lombok.NonNull;
 import lombok.Value;
 
 @Value
 @Builder
-public class SearchResultDTO {
+public class SearchResultDto {
```
ewelinajozwik: Why not a record?

projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
tgreenwood: the only reason is my absent-mindedness :) 

I supposed to use ${application.basePath}, but probably it will never change and there parametrization is redundant. What do you think?

tgreenwood: but I made it as planned originally

projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
tgreenwood: missed the gpt-generated part, thanks, renamed

projects/saturn/src/main/java/io/fairspace/saturn/services/workspaces/WorkspaceService.java
tgreenwood: It shouldn't be there for sure as well as the added try-catch. I rolled it back. 

There was a local debugger's glitch - something did not work, but when I was trying to debug it got stuck.

projects/saturn/src/main/java/io/fairspace/saturn/services/workspaces/WorkspaceService.java
tgreenwood: I rolled it back, it was a local debugger glitch, I was trying to catch a problem, but it was not here. Then I forgot to delete the auxiliary try-catch

projects/saturn/src/main/resources/application.yaml
tgreenwood: done, parametrized all controllers

relevance is to be discussed

projects/saturn/src/main/java/io/fairspace/saturn/controller/UserController.java
tgreenwood: good point, aligned with one style

projects/saturn/src/main/java/io/fairspace/saturn/controller/dto/SearchResultDto.java
tgreenwood: Wow, turned out there is an Intellij issue, while I thought @Builder is not applicable to records. Switched them to records too

projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
ewelinajozwik: true, we could just have "/api" everywhere, not sure why it was made configurable, but let's keep the configuration option then

projects/saturn/src/main/java/io/fairspace/saturn/controller/enums/CustomMediaType.java
```diff
@@ -0,0 +1,13 @@
+package io.fairspace.saturn.controller.enums;
+
+public enum CustomMediaType {
+    ;
```
ewelinajozwik, line 4:
```suggestion
```

tgreenwood: it is a must, otherwise enum would not be possible to compile

projects/saturn/src/main/java/io/fairspace/saturn/controller/ViewController.java
tgreenwood: pushed another approach setting it in one place at servlet level

PR #1580, opened 2025-01-02
Add resolvable URL query params filters, fix issues.

ewelinajozwik:
Adds feature to enable redirection links to Fairspace with metadata prefiltering as a first step for enhancing external tools integration with Fairspace.

Fixes smaller issues, including right panel property ordering using `sh:order`, metadata validation error message improvements and changes to metadata query filters to be case insensitive.


projects/mercury/src/metadata/views/MetadataViewContext.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+// import {useHistory} from 'react-router-dom';
```
peyman-mohtashami: Do we want to deliberately keep these comments in the code?

peyman-mohtashami (approved):
LGTM.
I reviewed just Reactjs-related files and not Java files.

projects/mercury/src/metadata/views/MetadataViewContext.js
ewelinajozwik: No, thanks for spotting that, those are removed now

PR #1582, opened 2025-02-25
Add cleanup of ThreadLocal for reusable threads from the pool.

ewelinajozwik:
ThreadLocal variable is used in Saturn RequestContext to store user IRI within a thread.

When using thread pools the threads are reused, so the user IRI used previously for a thread was preserved and was reused for a different request and often different user, causing invalid request user being handled. 

This PR adds ThreadLocal variables cleanup before each new request using request filters.

See: https://www.baeldung.com/java-threadlocal#threadlocalsand-thread-pools


pvannierop (approved):
Nice catch! LGTM

